PR #55, opened 2024-08-05
Image Generation Capability

mborish:
Necessary changes to generate binary images with specific characteristics instead of gcode.
@cadkin @liamnwhite1 @aroschli Don't know everyone's notification level.


src/managers/session_manager.cpp
```diff
@@ -233,13 +234,13 @@ namespace ORNL
         QString file_name;
         if(filename == "") // filename is blank because the part is being loaded via project import
         {
-            file_name = new_part->name();
```
aroschli, line 236: These changes cause a crash when loading a project. Simply reverting back to the previous version for these lines of code resolves the issue.

How to test:
Open Slicer 2
Load an object, slice it, save project
Close Slicer 2
Load the previously loaded project
It will crash with the code in its current state, but reverting to the previous version should work.

aroschli (changes requested):
Everything looks good and functions properly, just need to revert some of the changes to file_name in session_manager.cpp to prevent the project system from crashing.

src/managers/session_manager.cpp
liamnwhite1: Confirmed that the program crashes when loading a project.

liamnwhite1 (changes requested):
Program crashing when loading a project needs to be addressed.

aroschli (approved):

liamnwhite1 (approved):

PR #56, opened 2024-08-05
Use adjusted time for g-code output

aroschli:


src/threading/gcode_loader.cpp
```diff
@@ -249,6 +249,9 @@ namespace ORNL
                 Volume total_volume;
                 min_time = INT_MAX;
                 max_time = INT_MIN;
+                adjusted_min_time = INT_MAX;
+                adjusted_max_time = INT_MIN;
```
liamnwhite1, line 253:
```suggestion
                min_time = std::numeric_limits<int>::max();
                max_time = std::numeric_limits<int>::min();
                adjusted_min_time = std::numeric_limits<int>::max();
                adjusted_max_time = std::numeric_limits<int>::min();
```

liamnwhite1 (changes requested):
Looks good. My only recommendation is to use std::numeric_limits<int> instead of the INT_MAX/MIN macros.

liamnwhite1 (approved):

mborish (approved):

cadkin (approved):
lgtm!

PR #72, opened 2025-03-18
Fix/part relative centering

### LINKED ISSUE #37: "Center Part in Build Area" is malfunctioning.
## Expected Behavior
Self-Explanatory.

## Actual Behavior
Self-Explanatory.

## Possible Solution
TBD

Here I have loaded in multiple STLs that need to stay in relative position to each other. 
image

I would normally now select them all in the "Currently Manipulating" tree area .
image

Once, selected I can drag them around on XY plane and they hold their grouping just fine, but the shortcut to center them in the build volume actually breaks it and separates some of the STLs now. 
image

## Specifications
This was observed on pre-release version of V0.99.10

liamnwhite1:
Closes #37

### DIFF
--- a/include/graphics/view/part_view.h
+++ b/include/graphics/view/part_view.h
@@ -70,7 +70,12 @@ class PartView : public BaseView {
     //! \todo This should be done using the selected parts instead.
     void centerPart(QString name);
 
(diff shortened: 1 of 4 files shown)


aroschli (approved):
Centering appears to be working as expected

PR #83, opened 2025-05-29
fix: update normal vector to track slicing vector instead of fixed z-axis

### LINKED ISSUE #82: Bead Visualization - Normal to Print Vector
## Expected Behavior
Beads are shown in the visualization as normal to the print plane (slicing vector).  

## Actual Behavior
Some beads are angled, while others are still shown as "straight-down" or "3-axis" beads.  This makes it hard to tell if the print is consolidated or gapped, because the beads should be shown as normal to the slicing vector.  

## Possible Solution
Update visualization to always draw the beads as perpendicular to the slicing vector. 

Slice something at an angle and look at the visualization with true bead widths turned on.

See attached image for example:
Image

liamnwhite1:
Closes #82 

Fixes issue that cause incorrect visualization of gcode cylinders.

### DIFF
--- a/src/graphics/support/shape_factory.cpp
+++ b/src/graphics/support/shape_factory.cpp
@@ -1,1521 +1,1784 @@
-//Qt
-#include <QMatrix4x4>
-#include <QtMath>
 
-//System
-#include <vector>
-#include <math.h>
-
-//Local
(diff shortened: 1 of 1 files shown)


aroschli (approved):
Visualization looks great

PR #120, opened 2026-04-24
Add extrusion rate scaling for Ingersoll

### LINKED ISSUE #12: Corner Ramping Based on Path Type
## Expected Behavior

## Actual Behavior

## Possible Solution

Self-Explanatory.

## Specifications
N/A

aroschli:


liamnwhite1 (approved):
I made some minor changes to improve the performance but everything looks good on my end. When the CI checks have finished I'll rebase and merge it.

PR #127, opened 2026-05-18
fix(preferences): seed visualization color defaults safely

liamnwhite1:
## Summary

Seed visualization color defaults when preferences are created and safely load persisted color overrides.

## Related issues

- No related issues.

## Details

This fixes a fresh-install issue where visualization colors could remain uninitialized until first slicing, allowing invalid `QColor` values to be persisted as black.

The preferences manager now initializes all visualization colors from the enum defaults during construction. Color lookup also falls back to the enum default if a key is missing. Persisted visualization color values are parsed and applied only when valid; missing, empty, or invalid values remain at their defaults so they can be repaired on export.

The fix also marks visualization color edits and resets dirty so user changes are saved. Valid persisted black colors are preserved instead of being treated as corruption.

## Impact

Fresh installs now have valid visualization colors before slicing or exporting preferences.

Existing preferences continue to load valid visualization color overrides, including intentional black selections. Invalid or empty persisted color values are repaired by falling back to defaults.

Risk level: Low. The change is scoped to preference initialization and visualization color persistence.

## Testing strategy

- Completed a fresh install and slice on Windows 11.

### DIFF
--- a/include/managers/preferences_manager.h
+++ b/include/managers/preferences_manager.h
@@ -428,8 +428,11 @@ class PreferencesManager : public QObject {
     //! \brief Constructor
     PreferencesManager();
 
-    //! \brief Set Visualization Colors to Default
-    void setDefaultVisualizationColors(std::unordered_map<std::string, std::string> visualizationColorsHex);
+    //! \brief Set visualization colors to defaults and apply valid persisted overrides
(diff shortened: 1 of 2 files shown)


aroschli (approved):
Building and running correctly

PR #153, opened 2026-06-02
[codex] Restore versioned Windows installer artifact

### LINKED ISSUE #133: Add Run/Version Number to EXE Generations on GitHub Actions
## Summary
Each .exe generated by a GitHub action has the same name. It would be good to add a unique identifier to each .exe file so that installers can be kept separate.

## Background
We previously had this functionality, but it got broken.

liamnwhite1:
## Summary
- Restore the Windows installer artifact name to use the Nix derivation name.
- Keep the NSIS output filename aligned with the uploaded installer artifact name.

## Root Cause
The Windows CI job used to derive the installer filename from the Nix package name, which includes the application/version metadata. During the ORNLSlicer branding refactor, that was replaced with the fixed filename `ornlslicer-installer.exe`, causing every generated installer to share the same name.

## Impact
Installer artifacts are versioned again, for example `ornlslicer-x86_64-w64-mingw32-2.0.0-installer.exe`.

Closes #133.

## Validation
- `git diff --check origin/develop..HEAD`
- YAML parse of `.github/workflows/ci.yml`
- `nix eval --raw .#windows.ornl.ornlslicer.name`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
       - uses: cachix/install-nix-action@v27
       - name: Setup Environment
         run: |
-          echo "install_artifact=ornlslicer-installer.exe" >> $GITHUB_ENV
+          echo "install_artifact=$(nix eval --raw .#${{ env.ci_derivation }}.name)-installer.exe" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


aroschli (approved):
This is working correctly

PR #155, opened 2026-06-03
Rename Marlin Pellet syntax to JuggerBot

aroschli:
## Summary
- Rename the Marlin Pellet syntax enum/constant entries to JuggerBot while preserving syntax value 11
- Rename the Marlin Pellet writer files/classes to JuggerBotWriter
- Update settings, master config, loader, and slicer writer selection text/references

## Verification
- `git diff --check`
- `cmake -S . -B build/generic-llvm-ninja`
- `cmake --build build/generic-llvm-ninja --config Debug`


liamnwhite1:
Will be revisited.

PR #156, opened 2026-06-03
Add hexagonal prism shape generation

aroschli:
## Summary

Adds a generated hexagonal prism shape to the shape menu. The new action prompts for hexagon side length, prism height, and object name, then adds the generated closed mesh to the current session.

The mesh factory now exposes a dedicated hexagonal-prism generator type so saved sessions can reload generated hexagonal prisms instead of treating them as generic imported meshes.

## User Impact

Users can create regular hexagonal prism geometry directly from the UI alongside the existing rectangular prism, open-top prism, pyramid, cylinder, and cone options.

## Validation

- `git diff --check`


liamnwhite1 (approved):
Looks good to me! I added a commit to format the codebase as well.

PR #159, opened 2026-06-04
feat: import STEP files

aroschli:
## Summary
- add OpenCASCADE-based STEP/STP import that tessellates CAD geometry into the existing triangulated mesh workflow
- extend GUI and CLI model-file filters/validation to accept STEP and STP files
- add OpenCASCADE to CMake and Nix build dependencies

## Verification
- `nix develop -c cmake --preset generic-llvm-ninja`
- `nix develop -c cmake --build --preset debug-generic-llvm-ninja --target ornlslicer`
- `nix develop -c ./build/generic-llvm-ninja/Debug/ornlslicer --help`

Note: no STEP fixture is present in the repo, so this has not been smoke-tested with a real STEP asset.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,7 @@ function(FindDependencies)
     find_package(Boost         REQUIRED COMPONENTS system)
     find_package(CGAL          REQUIRED)
     find_package(nlohmann_json REQUIRED)
+    find_package(OpenCASCADE REQUIRED)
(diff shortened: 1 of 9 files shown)


liamnwhite1 (approved):
Everything looks good to me but we should keep an eye on this because OpenCascade is a massive dep to drag along just for this purpose. I think there are ways to disable modules we aren't using but we can look into this in a future issue.

PR #161, opened 2026-06-10
Fix: STL loading and part visualization

aroschli:
## Summary
These changes improve the visualization of STL files to make edges and other features more noticeable. Changes were also made to STL loading the clear up some errors and crashes with particular geometries.
- improve part-view lighting and add feature-edge rendering while suppressing flat triangulation edges
- make CGAL mesh repair resilient by falling back to the original mesh on repair failures
- skip CGAL hole filling for non-manifold boundary meshes to avoid warnings/assertions when loading problematic STLs

## Verification
- `nix develop -c cmake --build --preset debug-generic-llvm-ninja --target ornlslicer`
- `timeout 25s build/generic-llvm-ninja/Debug/ornlslicer --input_global_settings templates/Desktop_04mm.s2c --input_stl_files /home/rhc/develop/Chair.STL --output_location /tmp/ornl-chair-test.gcode --shift_parts_on_load false --align_parts false --single_slice_height 1`
- `rg -n "CGAL|assertion|check violation|Non-manifold|non-manifold|Skipping hole" /tmp/ornl-chair-load.log`
- `git diff --check`

### DIFF
--- a/include/geometry/mesh/closed_mesh.h
+++ b/include/geometry/mesh/closed_mesh.h
@@ -153,9 +153,13 @@ class ClosedMesh : public MeshBase {
     static std::pair<QVector<MeshVertex>, QVector<MeshFace>>
     FacesAndVerticesFromPolyhedron(MeshTypes::Polyhedron& mesh);
 
-    //! \brief cleans a supplied polyhedron
-    //! \param polyhedron the object to clean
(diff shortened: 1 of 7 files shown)


liamnwhite1 (approved):
These changes look good! I added a fix for a transparency bug I found while testing.

PR #163, opened 2026-06-16
Rotate g-code info arrow with camera

aroschli:
## Summary

- Rotate the bead / segment info direction arrow using the active G-code view matrix.
- Refresh the info overlay after camera view changes, mouse rotation, zoom, orthographic changes, and resets.
- Update the direction tooltip to reflect that the arrow now matches the current G-code view.

## Why

The segment info arrow previously used only the unrotated top-down XY angle, so it became misleading after rotating the camera. The displayed travel direction now follows the same camera orientation as the selected toolpath.

## Validation

- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer -j 2`


liamnwhite1 (approved):
Everything looks good on my end!

PR #164, opened 2026-06-17
Add smoothing algorithm options

aroschli:
## Summary
- Add a profile setting for selecting the smoothing algorithm.
- Add `SmoothingType` enum values for Douglas Peucker, Radial Distance, Perpendicular Distance, and Reumann-Witkam.
- Dispatch polygon smoothing through the selected psimpl algorithm while preserving Douglas Peucker as the default/fallback behavior.

## Impact
Users can choose the smoothing strategy from Profile > Special Modes when smoothing is enabled.

## Validation
- `cmake --build --preset debug-generic-llvm-ninja`

### DIFF
--- a/include/utilities/enums.h
+++ b/include/utilities/enums.h
@@ -509,6 +509,17 @@ inline QString toString(PathModifiers modifier_type) {
 }
 
 /*!
+ * \enum  SmoothingType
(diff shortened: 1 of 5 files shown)


liamnwhite1 (approved):
Looks good!

PR #169, opened 2026-06-19
Rename PolymerSlicer to PlanarSlicer

aroschli:
## Summary

Renames the planar slicing implementation from `PolymerSlicer` to `PlanarSlicer` across the codebase.

## Changes

- Renamed `polymer_slicer.{h,cpp}` to `planar_slicer.{h,cpp}`.
- Updated class names, constructor definitions, include paths, and session manager construction sites.
- Renamed the default slicer enum value from `kPolymerSlice` to `kPlanarSlice` while preserving numeric value `0` for existing settings compatibility.
- Updated the user-facing slicer option label from `Polymer` to `Planar` in canonical settings metadata and generated master config.

## Impact

This keeps the default planar slicing path behaviorally equivalent while making the naming more accurate. Existing persisted slicer values continue to map to the same numeric setting.

## Validation

- `nix develop --command cmake --preset generic-llvm-ninja`

### DIFF
--- a/include/managers/session_manager.h
+++ b/include/managers/session_manager.h
@@ -176,7 +176,7 @@ class SessionManager : public QObject {
     bool isBuildMode();
 
     //! \brief Signals the internal slicing thread to begin computation.
-    //! \note If the internal slicing thread is unset, this function assumes that it should be a polymer slice.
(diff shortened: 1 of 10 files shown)


liamnwhite1 (approved):
Everything looks good. I amended the commit with a comment update in the session manager to reflect the name change but that was it.

PR #184, opened 2026-07-12
Address feedrate scaling for path modifiers and improve time estimation

### LINKED ISSUE #5: Remove Feedrate Adjustment from Path Modifiers
## Expected Behavior
Feedrate needs adjusted during printing paths to meet min/max desired time, but path modifiers such as tip wipes don't need to be adjusted because that could break their functionality.

## Actual Behavior
Feedrate gets adjusted for all G1 commands

## Possible Solution
Self-Explanatory

Self-Explanatory. 

## Specifications
All platforms

aroschli:
## Summary
- Adds material settings to disable feedrate scaling for start-up, slow down, tip wipe, and spiral lift path modifiers.
- Skips disabled modifier paths during feedrate rewrite and min/max modifier calculations.
- Updates adjusted layer-time estimation so mixed scaled and unscaled moves use the corrected scalable-time bucket.

## Root Cause
Feedrate adjustment was applied at the layer level, but some modifier paths need to retain their original feedrate. The time estimate also assumed the whole layer shared one feedrate modifier, and some non-build modifier moves were not included in the adjustable-time bucket even when their `F` values changed.

## Impact
When modifier feedrate scaling is disabled, the generated G-code keeps those modifier moves at their intended speed while the remaining eligible paths can still be scaled. The adjusted time estimate now reflects faster unscaled modifier moves instead of remaining unchanged.

## Validation
- `nix develop -c ninja -C build/generic-llvm-ninja`
- `git diff --check`

Closes #5.


liamnwhite1:
## Review findings addressed

1. **Protected modifier paths could inherit scaled modal feedrates**  
   Feedrates are modal, so skipping a tagged command was insufficient when it omitted a redundant `F` value. The parser now tracks authored modal feedrates and inserts explicit transitions between scaled and protected spans, including across layer boundaries.

2. **Curved motions did not update modal feedrate state**  
   `F` values on G2, G3, and G5 commands were not reflected in scaling eligibility for subsequent motions. These commands now update the shared modal feedrate state using the correct velocity units.

3. **Scaling opt-outs were inaccessible with locally enabled modifiers**  
   The global start-up, slow-down, tip-wipe, and spiral-lift exclusions depended on locally overridable path toggles. They now depend on minimum-layer-time feedrate modification itself, keeping them available regardless of local path overrides.

4. **Adjusted-time estimates ignored per-command feedrate clamps**  
   Reported times used the requested layer modifier even when extrusion limits produced a different effective modifier. Adjustable time is now tracked per motion and calculated from the feedrate actually emitted.

5. **Settings footers did not round-trip**  
   The writer emitted the settings array as a single `0 {...}` entry, which the parser ignored. It now emits individual `key value` comments, allowing the new boolean flags and other required settings to reload correctly.

Each fix was committed separately.

liamnwhite1 (approved):
Everything looks good on my end. I made some changes more so related to other things while I was reviewing the files. Just make sure they didn't change the intent of this PR and then I think we can merge!

PR #185, opened 2026-07-12
Highlight settings that differ from Global

### LINKED ISSUE #78: Feature Request - Highlight Settings that are not set by global
When using STL-specific settings, or Layer specific settings, or settings mesh modify settings - there is no way for the user to tell which settings have been set by those methods or are still controlled by global settings.

If a user has many layer groups, models, and changes many settings, it is very easy to forget which settings were set by specific settings or which are still controlled by global settings.  Also if opening a previous project from several months ago, it is near impossible to remember this information.  

Suggested solution: when an STL is selected, OR a settings mesh is selected, OR a layer group is selected, the settings in the right hand panel would automatically change color or displays some icon, denoting that those settings were set specifically in that control (and are no longer controlled globally).

Example image:

Image

aroschli:
## Summary

- Highlight settings that are locally overridden with an accent-colored, bold label.
- Add a reset control that removes the selected local override and restores the global value.
- Remove redundant local overrides when their values match Global, including all components of vector settings.

## Why

STL-specific, layer-specific, and settings-mesh overrides were visually indistinguishable from values inherited from Global. This made local configuration difficult to audit and reset.

## Impact

Users can now identify local overrides at a glance and return an overridden setting to its Global value directly from the settings row.

## Validation

- `nix develop -c ninja -C build/generic-llvm-ninja`
- `git diff --check`

Closes #78

### DIFF
--- a/include/widgets/layerbar.h
+++ b/include/widgets/layerbar.h
@@ -72,7 +72,9 @@ class LayerBar : public QWidget {
   signals:
     //! \brief Signal that the selection has been altered.
     //! \param name_and_bases: Label plus list of ranges currently selected
-    void setSelectedSettings(QPair<QString, QList<QSharedPointer<SettingsBase>>> name_and_bases);
+    //! \param inherited_bases: Parent settings for each selected base; null entries inherit Global directly
+    void setSelectedSettings(QPair<QString, QList<QSharedPointer<SettingsBase>>> name_and_bases,
+                             QList<QSharedPointer<SettingsBase>> inherited_bases);
 
     //! \brief Signal that the selected layer settings ranges changed.
     //! \param part: Part that owns the selected ranges, or null when no layer range is selected.
--- a/include/widgets/settings/setting_bar.h
+++ b/include/widgets/settings/setting_bar.h
@@ -95,8 +95,10 @@ class SettingBar : public QWidget {
     void filter(QString str);
 
     //!\brief Currently selected settings bases as provided
-    //! \param settings_bases: List of currently selected ranges
-    void settingsBasesSelected(QPair<QString, QList<QSharedPointer<SettingsBase>>> name_and_bases);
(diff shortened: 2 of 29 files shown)


liamnwhite1:
## Findings and fixes

1. **Layer-range reset used the wrong inheritance model**

   **Finding:** The UI compared every local setting directly with Global. Layer ranges actually inherit settings through `Global → part → layer range`. Removing a range key could therefore make slicing inherit the part value while the UI incorrectly displayed Global. A range value equal to Global could also be removed even when it was needed to override the part.

   **Fix:** Selected settings now carry their inherited parent through the settings UI. Display, consistency checks, reset behavior, and redundant-value cleanup resolve values through the selected base, its parent, and then Global. Reset is now described as returning to the inherited value.

2. **Reloading settings modified project data without change tracking**

   **Finding:** Selecting an item or reloading settings could automatically delete an equal-to-Global local key. These reload paths bypassed the normal undo snapshot and project-modified notifications, so the change was neither undoable nor reliably persisted.

   **Fix:** Reload operations are now read-only. Redundant keys are removed only during explicit edits, where the existing about-to-change, undo, modified-state, and persistence paths are active. Override indicators compare values without mutating stored settings.

3. **Resetting inconsistent selections left stale warning indicators**

   **Finding:** Resetting multiple items with different values made the row consistent, but the tab and pane warning counts were not decremented. Repeated reloads could also inflate the warning count.

   **Fix:** Warning updates now represent transitions between consistent and inconsistent states. Selection changes account for counters that were already reset, while reset, repeated reload, and undo/redo operations emit balanced warning deltas.

### Commits

- `731af834` — `fix(settings): respect inherited override values`
- `d7906f43` — `fix(settings): balance warning counts across reloads`
- `afded8c6` — `fix(settings): compare override indicators by value`

liamnwhite1 (approved):
Everything looks good. Added a few changes to account for settings regressions when reset.

PR #186, opened 2026-07-12
Fix infill path linking for closest path optimization

### LINKED ISSUE #112: [BUG] Infill Closest Path Optimization
## Describe the bug

Line infill is made of parallel toolpaths. Minimum travel distance is used to define if these toolpaths are connected with travels or extrusion moves. If extrusion moves are used, the end points are connected using the next closest point strategy. This means segments alternate from printing right to left and left to right. If travels are used, the connections are generated such that all paths are either left to right or right to left. The starting point for each segment is always on the same end.

It is desirable to maintain the current behavior when travels are used but also have the option to implement some sort of next closest point approach so that paths alternate between left to right and right to left.

## To Reproduce
Project for disconnected infill attached, connected infill, and a workaround with skeletons that achieves the desired behavior by using a different model geometry.

## Expected behavior
Ability to print the line infill pattern with and without travels as connections. With travels, have the ability to control which end of each line is the starting point.
## Screenshots

Infill with travels:
 
Image

Infill with connecting extrusion segments:
 
Image

Desired behavior forced with Skeletons:
 
Image

[Project Files.zip](https://github.com/user-attachments/files/26445875/Project.Files.zip)

aroschli:
## Summary

Fixes #112 by making the existing line-infill linking option allow travel-separated line infill to choose the next path from all remaining line endpoints instead of only the monotonic front/back candidates.

## What changed

- Keeps the default monotonic behavior when `infill_lines_partitioned_linking` is disabled.
- Uses the selected Path Order Optimization when `infill_lines_partitioned_linking` is enabled, allowing closest-endpoint travel ordering so adjacent disconnected lines can alternate direction.
- Uses the live current location for non-custom Point Order Optimization in the opt-in path, preventing default-origin endpoint selection from overriding the closest-end decision.
- Updates the setting display text and tooltip, and regenerates `master.conf`.

## Impact

Existing profiles remain compatible because the stored setting key is unchanged. Users can now enable the line infill option to get closest-path behavior for disconnected line infill while preserving the old monotonic travel ordering by leaving it disabled.

## Validation

- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_obj --parallel 2`
- `git diff --check`


liamnwhite1:
| Review finding | Associated fix |
| -- | -- |
Next Farthest selected the globally farthest endpoint instead of the farthest path. This could select a line whose usable entry point was actually nearby. | Candidates are now ranked using each line’s nearest endpoint distance. Path Order selects the line, while Point Order independently chooses its orientation.
The Lines-only option could affect other infill patterns. A stored enabled value remained active after switching to Grid, Triangles, or Honeycomb. | setInfillParameters() now enables the behavior only when the active pattern is InfillPatterns::kLines.
Outside In and Inside Out used raw first/last vector entries. This did not produce genuine exterior-to-center or center-to-exterior traversal. | Outside In alternates between the remaining exterior scanlines toward the center. Inside Out repeatedly selects the middle remaining scanline and progresses outward. State resets for every geometry batch.
The setting name implied exclusively Next Closest behavior. Runtime behavior actually follows the configured Path and Point Order optimizations. | Renamed the control to “Apply Path and Point Ordering to Infill Lines”, clarified its tooltip, and regenerated master.conf from the canonical YAML.

</body></html>

PR #187, opened 2026-07-12
Localize initial extruder speed

### LINKED ISSUE #58: Initial Extruder Speed - Layer and Part Specific Settings
## Expected Behavior
The ability to adjust initial extruder speed on a part or layer specific basis.

## Actual Behavior
Only a global value is passed to the writer.

## Possible Solution
Self-Explanatory

Self-Explanatory.  Ideally, bullet or numbered list.

## Specifications
Easily tested with Cincinnati Syntax

aroschli:
## Summary

Fixes #58 by carrying initial extruder speed through segment-local settings and having writer startup logic prefer the segment value before falling back to the global machine setting.

## Root Cause

`MS::Extruder::kInitialSpeed` could be overridden at the part or layer setting level, but generated segments and writer `writeExtruderOn` calls generally read the global settings bag. That meant Cincinnati and other writers with initial-RPM startup logic could not honor part- or layer-specific initial speed values.

## Changes

- Add a shared writer helper for segment-local initial extruder speed lookup with global fallback.
- Thread segment settings into relevant `writeExtruderOn` calls across writers that already implement initial-speed behavior.
- Preserve initial extruder speed on generated travel, lead-in, ramp, startup, slowdown, and split segments.
- Stamp initial extruder speed into generated region segments for infill, skin, support, raft, brim, and skirt paths.

## Impact

Part- and layer-specific initial extruder speed overrides now reach the G-code writer startup path for Cincinnati and the other writers that support initial extruder speed, without changing behavior for writers that do not currently implement initial-RPM startup syntax.

## Validation

- `git diff --check`
- `git diff --cached --check`
- `nix develop -c cmake --build --preset debug-generic-llvm-ninja -j 2`

### DIFF
--- a/include/gcode/writers/aml3d_writer.h
+++ b/include/gcode/writers/aml3d_writer.h
@@ -93,7 +93,8 @@ class AML3DWriter : public WriterBase {
     //! \brief Writes G-Code to disable the tamper
     QString writeTamperOff();
     //! \brief Writes G-Code to enable the extruder
-    QString writeExtruderOn(RegionType type, int rpm, int extruder_number);
+    QString writeExtruderOn(RegionType type, int rpm, int extruder_number,
+                            const QSharedPointer<SettingsBase>& params = nullptr);
     //! \brief Writes G-Code to disable the extruder
     QString writeExtruderOff(int extruder_number);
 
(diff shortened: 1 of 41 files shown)


liamnwhite1 (approved):
Looks good to me!

PR #188, opened 2026-07-12
Layer height changes for layer specific settings

aroschli:
## Summary

Fixes layer-specific settings ranges when a layer-height override reduces the total layer count.

- Clamp out-of-bounds range endpoints to the new final valid layer instead of deleting the endpoint and collapsing the range unexpectedly.
- Recalculate the layer count immediately after manually editing a layer range, so changing the start of a larger-height range cannot leave a stale high endpoint selected.
- Make settings range lookup and limit remapping more defensive, and keep template/user range metadata in sync when ranges are removed or remapped.

## Root Cause

Layer-height changes can reduce the number of valid layers. The layer bar previously deleted dots whose layer index no longer existed, which turned a two-ended range into a single-layer range. A follow-up manual range edit could move the override earlier, reducing the total layer count again while leaving stale visual/model state behind.

## Validation

- `git diff --check`
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer`

### DIFF
--- a/include/widgets/layerbar.h
+++ b/include/widgets/layerbar.h
@@ -249,6 +249,9 @@ class LayerBar : public QWidget {
     //! \brief Move dot to layer specified.
     bool moveDotToLayer(LayerDot* dot, int layer);
 
+    //! \brief Moves an out-of-bounds range endpoint to the new top layer when possible.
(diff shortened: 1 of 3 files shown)


liamnwhite1 (approved):
These look great!

PR #189, opened 2026-07-12
Optimization Point Improvements

aroschli:
## Summary

- Show custom optimization point markers in the G-code toolpath view using the existing toolbar toggle.
- Place optimization point markers on the printer bed in both part and G-code coordinate modes.
- Allow visible optimization point markers to be clicked and dragged in either view, updating the associated X/Y settings as they move.
- Refresh optimization marker visibility and position when relevant optimization settings change.

## Impact

Users can inspect and adjust custom island, path, and point optimization locations directly from the 3D views instead of manually editing coordinate settings. The same toolbar button now controls marker visibility consistently across Part View and G-Code View.

## Validation

### DIFF
--- a/include/graphics/objects/printer/printer_object.h
+++ b/include/graphics/objects/printer/printer_object.h
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <QMatrix4x4>
+#include <QPointF>
+#include <QString>
 #include <qlist.h>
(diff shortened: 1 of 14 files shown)


liamnwhite1 (approved):
This looks great! I've wanted this for so long!!

PR #192, opened 2026-07-13
Support Structures Reimplementation

aroschli:
## Summary

- Reimplement planar support detection and propagation so overhangs and objects above the build plate receive support.
- Generate conventional grid supports by default, with optional organic/tree branches.
- Add inward hollow tapering with configurable angle and minimum tube-wall contours.
- Add configurable dense interface layers, interface expansion, conventional wall contours, and expanded solid base layers.
- Add bridge-aware suppression, Everywhere/Build Plate Only placement, settings-mesh blockers and enforcers, and connectivity validation.
- Link support grid and interface hatch lines with extrusion moves instead of producing many disconnected paths.
- Persist the new settings in the canonical metadata, generated configuration, and shipped templates.

## Root cause

The previous support pipeline coupled overhang detection and downward propagation to assumptions about model placement, so support demand could be missed or fail to reach a valid foundation. Tapering and dense support regions were also partitioned inconsistently, and hatch lines were emitted as separate paths connected by travel moves.

## Impact

Support generation is substantially more predictable and configurable for the planar slicer. The implementation is intentionally limited to planar slicing and does not enable these behaviors in radial, conformal, or other slicers.

## Validation

- Built the Debug executable with:
  `nix develop -c cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer -j8`
- Verified support generation for a model positioned 20 mm above the bed.
- Verified interface expansion, three solid base layers, base expansion, and configurable support wall contours.
- Verified bridge suppression removes support for an anchored span within the configured bridge limit.
- Verified Everywhere retains model-landed support while Build Plate Only removes it.
- Verified connectivity validation rejects support below the configured minimum base area.
- Parsed the generated master configuration and all shipped templates as JSON.
- Ran `git diff --check`.


liamnwhite1 (approved):
These look great! I encountered a slowdown when it came to visualization time but I think this will probably be fixed by #197 .

Issue #35, opened 2024-01-25
Height for Angled Printing


aroschli:
## Expected Behavior
I believe the bottom of the object Z height should be printed at layer height / sin(slicing angle) + z offset

In the attached project, the value is 0.1914 whereas it should be 0.1414. This 0.1414 is calculated with the following: layer height 0.1, slicing angle 45, and z offset of 0.     0.1/sin(45°)  + 0 = 0.1414

## Actual Behavior
The Z height seems too large

## Possible Solution
Self-Explanatory

## Steps to Reproduce the Problem
Project attached.

## Specifications
N/A
[AngledSlicingHeightBug.zip](https://github.com/ORNLSlicer/Slicer-2/files/14055569/AngledSlicingHeightBug.zip)

aroschli:
This bug was reported when slicing plane was defined by angle, but it's now defined by angle.

This issue still persists.

aroschli:
Addressed with PR #193

PR #193, opened 2026-07-13
Fix angled slicing Z calcluations

aroschli:
## What changed
This updates layer reorientation for angled slicing so restored toolpaths are lifted to the minimum printable Z implied by the slicing plane instead of always relying on the horizontal half-layer-height adjustment.

## Why
Issue #35 showed that angled slices could start too low or too high because the post-rotation Z shift was based on the flattened layer center rather than the actual printable height of the reoriented extrusion paths. For a 0.1 mm layer at 45 degrees, the first printable Z should be about 0.1414 mm.

## Root cause
The old logic always applied the same half-layer-height style offset after reorientation. That works for horizontal slicing, but once the layer is rotated the minimum printable Z depends on the slicing plane normal and on the lowest material-depositing segment, not just the original layer center.

## Impact
Angled slicing now starts at the expected printable Z for the current plane orientation while preserving the existing horizontal and spiralize behavior.

## Validation
- `nix develop --command cmake --build build/generic-llvm-ninja --target ornlslicer -j16`
- `nix develop --command clang-format --dry-run --Werror include/step/layer/layer.h src/step/layer/layer.cpp`
- `git diff --check`
- Verified generated G-code first-print Z values for horizontal, 45 degree, 30 degree, 60 degree, and 45 degree plus Z-offset cases

### DIFF
--- a/include/step/layer/layer.h
+++ b/include/step/layer/layer.h
@@ -87,9 +87,24 @@ class Layer : public Step {
     QList<QSharedPointer<IslandBase>> m_island_order;
 
   private:
+    //! \brief Builds the translation used to move paths between the flattened slicing frame and printer coordinates.
+    Point getOrientationShift() const;
+
+    //! \brief Moves restored paths above the minimum printable Z height for the slicing plane.
+    void applyMinimumZShift();
+
+    //! \brief Removes the minimum printable Z shift before flattening paths again.
+    void removeMinimumZShift();
+
+    //! \brief Returns the minimum Z coordinate among material-depositing segments.
+    float getMinimumPrintZ();
+
     //! \brief Creates tree-like structure if brims exist, otherwise, sorts islands into precendence order
     QList<QHash<QSharedPointer<IslandBase>, QList<QSharedPointer<IslandBase>>>>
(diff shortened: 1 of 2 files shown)


aroschli:
Note, this PR fixes the issue with improper calculating layer height and position on the bed. This is separate from issues that can cause toolpaths (such as tip wipes) to collide with the bed.

liamnwhite1:
Found a bug where later normal layers are shifted half a layer too low
`src/step/layer/layer.cpp (line 325)` computes minimum_print_z as a constant layer_height / z_component, and `getOrientationShift() (line 298)` no longer adds the previous non-spiral half-layer shift. That only fixes/clamps the first layer. For horizontal slicing with 0.1 mm layer height, layer centers are 0.05, 0.15, 0.25, etc.; after this PR the first layer gets lifted from 0.05 to 0.1, but the second layer stays at 0.15 because 0.15 >= 0.1. Previously it would be 0.2.
applyMinimumZShift() now bases the target on the restored layer’s current minimum print Z plus the projected half-layer clearance, layer_height / (2 * abs(normal.z())), instead of clamping every layer to the same first-layer floor. It also skips the shift when there are no material-depositing segments.

liamnwhite1 (approved):
Looks good! Just added a fix for one minor bug.

Issue #22, opened 2023-12-22
Settings mesh disappears when trying to rotate a parent/child object


aroschli:
## Expected Behavior
When a build mesh and settings mesh are combined in a parent/child relationship, they translate as one. It is expected that they would also rotate as one also.

## Actual Behavior
When applying a rotation to the parent/child object, the settings mesh disappears from view. It does not seem possible to re-center it or bring it back into view. It is still listed on the window in the lower left.

## Possible Solution
Self-Explanatory

## Steps to Reproduce the Problem

1. Load build mesh and settings mesh
2. Set settings mesh as child of parent build mesh
3. Translate parent/child object to correct position
4. Attempt to rotate the same object

## Specifications
N/A

aroschli:
This issue has been resolved with PR #194

PR #194, opened 2026-07-13
Fix parent child rotation bug

aroschli:
## Summary
- keep rotation as a live transform for parts that participate in parent/child relationships
- avoid baking parented rotations into the mesh, which reset the part transform and break hierarchy propagation
- preserve the existing rotation-bake behavior for standalone parts

## Why
Issue #22 reported that when a settings mesh is parented to a build mesh, translating the group works but rotating it causes the settings mesh to disappear. The root cause was that `PartMetaItem::setRotation` baked rotations into the mesh and reset the live transform. That works for standalone parts, but it breaks parent/child transform propagation because the hierarchy needs the rotation to remain on the graphics object transform.

## Impact
Parent and child parts now rotate together as expected, including settings meshes attached to build meshes. Standalone parts keep their previous behavior.

## Validation
- Reviewed the parent/child transform flow from `PartMetaItem` through `PartView` and `GraphicsObject`
- Confirmed the branch only contains the issue #22 fix
- Build not run in this environment because `cmake` is unavailable

### DIFF
--- a/src/widgets/part_widget/model/part_meta_item.cpp
+++ b/src/widgets/part_widget/model/part_meta_item.cpp
@@ -117,7 +117,12 @@ void PartMetaItem::setRotation(QQuaternion r, bool current_rotation) {
     if (m_rotation == r)
         return;
 
-    if (current_rotation) {
+    const bool has_parenting = !m_parent.isNull() || !m_children.isEmpty();
+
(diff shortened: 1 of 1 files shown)


liamnwhite1 (approved):
The changes look good!

PR #195, opened 2026-07-13
Prevent bed collisions for tip wipes and spiral lifts during angled printing

### LINKED ISSUE #66: Tip wipe & bed collision with angled slicing
## Expected Behavior
When using tip wipes with angled slicing, no collisions between the tip wipes and the bed should occur.

## Actual Behavior
When using tip wipes with angled slicing, tip wipes are generated such that they collide with the bed.

## Possible Solution
If a collision is detected between a potential tip wipe and the bed, the tip wipe should be redirected such that it does not collide with the bed. To accomplish this, we can use a CGAL::Segment_3 to model the tip wipe, a CGAL::Plane_3 to model the bed, and CGAL::do_intersect(CGAL::Segment_3, CGAL::Plane_3) do detemine if there is a collision.

N/A

## Specifications
N/A

aroschli:
## Summary

This change prevents angled-printing clearance moves from intersecting the build plate after layer reorientation.

## What changed

- added a post-`reorient()` clearance pass in `Layer` that checks tip wipes and spiral lifts against the build plate plane in machine coordinates
- redirects colliding clearance moves by mirroring their in-plane motion while preserving lift along the slicing normal
- rewrites redirected arc-based spiral lifts as linear clearance moves so the emitted geometry stays valid after redirection
- removed the older angled-slicing override in infill and skin that forced tip wipes to `REVERSE TIP WIPE`

## Root cause

Tip wipes and spiral lifts were generated in slice-local coordinates. During angled printing, those moves could become bed-crossing motions only after the layer was rotated back into machine space, but there was no post-reorientation collision check to catch them.

## Impact

- forward tip wipes can now remain forward during angled slicing when requested
- tip wipes and spiral lifts that would otherwise cross `z = 0` are redirected to a safe path instead of colliding with the bed
- Cincinnati and other writers benefit from the same geometry fix because the correction happens before G-code emission

## Validation

- `nix develop -c ninja -C build/generic-llvm-ninja ornlslicer`
- verified emitted path modifiers still surface in G-code comments for practical inspection (`TIP WIPE`, `SPIRAL LIFT`)

## Issue

- fixes #66

### DIFF
--- a/include/step/layer/layer.h
+++ b/include/step/layer/layer.h
@@ -99,6 +99,9 @@ class Layer : public Step {
     //! \brief Returns the minimum Z coordinate among material-depositing segments.
     float getMinimumPrintZ();
 
+    //! \brief Redirects clearance moves that would intersect the build plate.
+    void redirectClearanceMoves();
+
     //! \brief Creates tree-like structure if brims exist, otherwise, sorts islands into precendence order
     QList<QHash<QSharedPointer<IslandBase>, QList<QSharedPointer<IslandBase>>>>
     createSequence(QList<QSharedPointer<IslandBase>> parent, QList<QList<QSharedPointer<IslandBase>>> children);
--- a/src/step/layer/layer.cpp
+++ b/src/step/layer/layer.cpp
@@ -4,6 +4,7 @@
 #include <cstdlib>
 #include <limits>
 
+#include <CGAL/intersections.h>
 #include <qcontainerfwd.h>
 #include <qhash.h>
 #include <qhashfunctions.h>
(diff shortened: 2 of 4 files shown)


liamnwhite1:
Findings and fixes:

1. The first clearance segment was checked from its stored segment start, but the writer emits movement from the previous endpoint to the segment endpoint. For generated tip wipes, that could miss the real machine move and allow a bed-crossing path through.

   Fix: `ad85c45a fix(layer): seed clearance checks from emitted path position`
   The clearance pass now seeds each clearance run from the previous segment endpoint when available, so the collision check matches emitted motion.

2. Arc spiral lifts were checked only as a straight chord from start to end. A near-full-circle arc could sweep below the build plate even when its endpoints, and therefore the chord, looked safe.

   Fix: `b5abcd8b fix(layer): check swept arc clearance against build plate`
   The clearance pass now detects `ArcSegment`s and samples across the emitted arc sweep with bounded angular steps and linear Z interpolation. Degenerate arcs fall back to the line check.

Issue #47, opened 2024-05-16
Final Z Lift


aroschli:
## Expected Behavior
Allow the user to define a distance to lift the extruder at the end of the print.

## Actual Behavior
Self-Explanatory

## Possible Solution
New setting for final lift distance

## Steps to Reproduce the Problem
Self-Explanatory.

## Specifications
Should be normal to the slicing plane.

aroschli:
User sets a distance, say 10in. At the conclusion of the print, before the writer's writeShutdown, the system issues a Z lift for 10in. This should be the final motion command.

aroschli:
Addressed in PR #196

PR #196, opened 2026-07-13
Add Final Z Lift Functionality

aroschli:
## Summary
This PR adds configurable final lift behavior at the end of a print.

## What changed
- Added a new `final_lift_distance` profile setting.
- Implemented Cincinnati final lift output before shutdown, following the slicing plane normal.
- Updated the final lift comment so the UI recognizes it as travel.
- Added shared writer-base tracking for the last emitted point and reusable lift-vector helpers.
- Extended final lift support to the traditional point-based writers, while intentionally skipping radial and other special-case writers for now.

## Why
Issue #47 requested a user-configurable end-of-print lift that occurs before shutdown and follows the slicing plane normal instead of assuming machine Z.

## Impact
- Users can configure a final lift distance as part of travel settings.
- Supported writers now emit a final travel-style lift before their normal shutdown sequence.
- The UI colors the Cincinnati final-lift comment correctly as travel.

## Validation
- Built successfully with `nix develop -c ninja -C build/generic-llvm-ninja`.

### DIFF
--- a/include/gcode/writers/cincinnati_writer.h
+++ b/include/gcode/writers/cincinnati_writer.h
@@ -96,6 +96,9 @@ class CincinnatiWriter : public WriterBase {
     //! \brief Writes G-Code to update the acceleration value
     QString writeAcceleration(Acceleration acc);
 
+    //! \brief Writes the end-of-print lift before shutdown logic executes.
+    QString writeFinalLift();
+
     //! \brief Writes gcode coordinates WXYZ for a move or travel to the destination point
     QString writeCoordinates(Point destination);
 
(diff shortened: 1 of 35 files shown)


liamnwhite1 (approved):
Found an issue where segment writers were recording m_end as the writer’s current output position even when the selected writer emitted no G-code for that segment. That could make final_lift_distance lift from a coordinate the machine never reached, especially for unsupported scan/arc paths.
Implemented a solution that updated line, arc, scan, and travel segment emission to call setCurrentPosition(m_end) only when the delegated writer returns non-empty output.

PR #197, opened 2026-07-14
G-Code Segment Visualization Improvements

aroschli:
## Summary

This PR improves G-code visualization performance and control for large segment counts while preserving quick true-width toggling.

## What changed

- Avoid expensive full-file true bead mesh construction when the estimated preview geometry exceeds the configured threshold.
- Add a configurable G-code preview mode preference: Auto, True Bead Widths, or Thin Lines.
- Add a configurable true-width vertex threshold in Preferences > Visualization.
- Keep the existing True Bead Widths checkbox as a quick on/off request for true-width rendering.
- For large G-code files, keep the full preview lightweight and build a true-width overlay only for the currently visible printable layer/segment range when it is under the threshold.
- Cache the visible-range overlay key so unchanged layer/segment windows do not rebuild unnecessarily.
- Reduce first-visualization stalls by preallocating preview buffers, avoiding empty text searches, and only reporting visualization complete after queued UI updates are issued.

## Why

Visualization was the final step of slicing and could appear stuck at 99-100%, especially on first load or when large objects triggered costly true-width mesh decisions. The old lightweight fallback also meant large files could not inspect true bead widths even for a small visible layer range.

## Impact

- Large G-code files should load into visualization using lightweight lines instead of freezing while constructing full true-width geometry.
- Users can tune the threshold that decides when true-width geometry is allowed.
- Users can still toggle true bead widths quickly from the G-code bar.
- Single-layer or small visible-range inspection can show true bead widths even when the full file is too large for full true-width rendering.

## Validation

- `nix develop --command build/generic-llvm-ninja/Debug/ornlslicer --input_project_file /home/rhc/develop/SpiralCylinder.s2p --output_location /tmp/SpiralCylinderVizTest`
- `git diff --check`

### DIFF
--- a/include/graphics/objects/gcode_object.h
+++ b/include/graphics/objects/gcode_object.h
@@ -33,8 +33,9 @@ class GCodeObject : public GraphicsObject {
     //! \param view: View to render to.
     //! \param gcode: GCode segments to visualize.
     //! \param segmentInfoControl: Segment / Bead info display control
+    //! \param use_true_widths: If true, draw printable moves as bead meshes when they are small enough to load.
     GCodeObject(BaseView* view, QVector<QVector<QSharedPointer<SegmentBase>>> gcode,
-                QSharedPointer<GCodeInfoControl> segmentInfoControl);
+                QSharedPointer<GCodeInfoControl> segmentInfoControl, bool use_true_widths);
 
     //! \brief Destructor.
     ~GCodeObject();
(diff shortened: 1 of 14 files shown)


liamnwhite1 (approved):
Review findings addressed:

1. The true-width overlay was rendered but not used for picking. Hover and click still hit-tested only against the lightweight centerline object, so selecting visible bead geometry could fail unless the cursor landed on the hidden centerline.
   - Fixed by making G-code picking prefer the active true-width overlay, then fall back to the primary lightweight object for travel moves and thin-line cases.

2. `Auto` and `True Bead Widths` preview modes behaved the same. Both used the configured vertex threshold, making the separate `True Bead Widths` option misleading.
   - Fixed by keeping `Auto` threshold-limited and making `True Bead Widths` explicitly bypass the automatic threshold fallback when the toolbar true-width toggle is enabled. `Thin Lines` still disables true-width preview generation.

PR #201, opened 2026-07-14
Set number of  object instances

aroschli:
## Summary
- Adds a right-click menu action for setting the number of instances for a selected object.
- Adds model-level instance counting, creation, and removal using the existing object copy naming pattern.
- Keeps copied part and mesh names aligned so session save/load paths see the expected names.

## Validation
- `nix develop -c cmake --build build/generic-llvm-ninja`
- `git diff --check`


liamnwhite1 (approved):
This looks great! Thank you for adding this last minute!

PR #203, opened 2026-07-14
Fix skin generation under perimeters

### LINKED ISSUE #202: Generate Skins Under Perimeters
## Summary
Top skins only generate under areas of no object (areas of upcoming "air"). The issue is that these upcoming voids need to be bounded by perimeters that need some way to be supported.

## Problem or opportunity
A perimeter printed on top of infill needs support. A skin makes sense, but the skin doesn't generate on the layers before the perimeter starts.

Layer of infill:

Image

Layer of perimeter on top that would fall into the infill:

Image

Project file is attached.

aroschli:
## Summary
- Add future-perimeter support geometry to top skin generation.
- Preserve zero-count behavior for top and bottom skin settings.

## Root Cause
Top skin only considered current-layer material that becomes air on upcoming layers, so perimeters introduced above sparse infill could start without a solid skin support region below them.

## Validation
- `git diff --check`
- `cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer`

Fixes #202

### DIFF
--- a/src/step/layer/regions/skin.cpp
+++ b/src/step/layer/regions/skin.cpp
@@ -55,6 +55,15 @@ PolygonList printableFootprint(const QVector<Polyline>& lines, Distance bead_wid
 
     return footprint;
 }
+
+PolygonList futurePerimeterSupport(const PolygonList& current_geometry, const PolygonList& upper_geometry,
+                                   Distance support_width) {
+    if (support_width <= 0 || upper_geometry.isEmpty())
+        return PolygonList();
+
+    PolygonList support_geometry = upper_geometry - upper_geometry.offset(-support_width);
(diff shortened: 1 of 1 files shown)


aroschli:
Follow-up changes pushed in `bc4f3c76`:

- Added comments clarifying the future-perimeter support band and the intentional `top_count == 0` behavior.
- Switched the skin geometry loops over `PolygonList` collections to `const PolygonList&` to avoid unnecessary copies.
- Re-ran validation:
  - `git diff --check`
  - `nix develop --command cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer`

I did not find an existing focused regression-test harness for this skin geometry path, so the added validation note is the lightweight artifact for this PR.

PR #204, opened 2026-07-14
Add minimum segment length settings

aroschli:
## Summary
- Add per-region minimum path segment length settings for perimeter, inset, skeleton, skin, infill, and support.
- Add `Polyline::removeShortSegments` to collapse short path segments before paths become G-code segments.
- Apply the cleanup during region path generation and regenerate the master settings config.

## Why
Smoothing the sliced polygon does not prevent later path generation, especially concentric offsets, from creating very short path segments. This adds a path-level cleanup control for each region type.

## Validation
- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja -j 2`

### DIFF
--- a/include/geometry/polyline.h
+++ b/include/geometry/polyline.h
@@ -95,6 +95,17 @@ class Polyline : public QVector<Point> {
      */
     Polyline simplify(Distance tol);
 
+    /*!
+     * \brief Collapses segments shorter than \p min_segment_length.
+     *
+     * \param min_segment_length Minimum allowed segment length. A value less than or equal to zero leaves the polyline
+     * unchanged.
+     * \param closed Whether to evaluate the segment from the final point back to the first point.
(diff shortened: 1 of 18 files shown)


aroschli:
Pushed follow-up fix commit `5e00b754`.

Changes made:
- Added an empty-path guard in spiralize perimeter optimization before removing the final segment. This prevents `removeLast()` from running after `removeShortSegments` causes `createPath` to return an empty `Path`.
- Updated skin path creation so the optimizer passes the actual pattern being processed. This keeps gradual skin using `skin_gradual_infill_pattern` semantics instead of always rereading `skin_pattern` when deciding whether the path is closed/concentric.

Validation:
- `git diff --check`
- Focused Ninja compile of `perimeter.cpp.o` and `skin.cpp.o`

I did not add a unit test because this repo does not appear to have an existing test target or test framework wired into CMake; I kept the fix scoped to the PR behavior.

PR #205, opened 2026-07-14
Adaptive bead width for perimeters and insets

aroschli:
## Summary

Adds adaptive bead-width support for perimeter and inset contour paths. The new settings allow the feature to be enabled independently for perimeters and insets, with minimum and maximum bead-width limits to preserve the existing fixed-width behavior when disabled.

## Details

- Adds perimeter and inset adaptive-width settings and generated config constants.
- Computes adaptive contour widths from the remaining printable shell area and applies the computed width to generated segments.
- Adjusts segment speed inversely with adapted bead width to preserve deposited area rate.
- Writes adapted contour widths into Cincinnati G-code comments and parses those widths back in the selected file distance unit.
- Updates G-code visualization so `AD-PERIMETER`, `AD-INSET`, and skeleton width comments render with their encoded bead widths.
- Handles hollow shell cases and contour-count transitions so curved walls do not collapse center paths onto nearly shared centerlines.

## Root Cause

Fixed-width perimeter and inset paths were placed only where nominal-width contours fit. Narrow shell gaps that were not an exact multiple of the nominal bead width could either leave unfilled regions or force fragile skeleton/inset behavior. Curved shell geometry also exposed edge cases where adaptive shell paths could be accepted after clamping, leaving thin residual geometry that produced overlapping center contours.

## Validation

- `cmake --build build/generic-llvm-ninja --config Debug --parallel 2`
- `git diff --check`


aroschli:
Addressed the review findings in `73f0ae46`:

- Propagated `SS::kAdapted` through modifier-generated segments so Cincinnati comments keep encoded adaptive widths on lead-in/startup/slowdown/tip-wipe/ramp/spiral segments.
- Made adaptive contour preview validation use the same short-segment cleanup as emitted path validation.
- Marked the adaptive perimeter/inset enable/min/max settings as non-local, matching the whole-region contour planning behavior.

Validation:

- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja --config Debug --parallel 2`

PR #207, opened 2026-07-14
Seam Attractor

aroschli:
## Summary

Adds a seam attractor for custom optimization points used during angled planar slicing.

## What changed

- Adds an `OptimizationAnchor` helper that projects custom island/path/point anchors onto the active slicing plane.
- Adds optional Z settings for custom island, path, point, and second point anchors while preserving existing X/Y defaults.
- Adds a grouped `Seam Attractor Vector` setting. A zero vector uses the slicing vector automatically; a nonzero vector gives the attractor its own projection direction.
- Converts projected anchors into the flattened optimization frame before island, path, and point ordering decisions.
- Threads optimization-frame context from layers into islands and regions.
- Updates seam marker visualization with a guide line that follows the active seam attractor vector.
- Refreshes seam guide visuals when seam anchor, attractor vector, or slicing vector settings change.
- Aligns slicing-plane visual refresh paths so part-view planes are updated when parts are shown, loaded, or settings change.

## Impact

For standard Z slicing, existing X/Y custom optimization behavior is preserved. For angled slicing, users can place a 3D seam attractor point and choose the vector used to project that attractor onto each slicing plane. Leaving the seam attractor vector at `{0, 0, 0}` preserves the automatic slicing-vector behavior.

## Validation

- `python3 scripts/generate_master_config.py resources/settings resources/configs/master.conf resources/configs/setting_inputs.conf`
- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug`


aroschli:
Pushed follow-up fix `a64d0b7a` for the review issues:

- Replaced stale `PS::Slicing::kSlicingVectorX/Y/Z` references with the current `kSlicePlaneNormalX/Y/Z` constants.
- Refreshed part-view optimization/seam guides when the slice plane normal changes, so the zero seam-attractor vector fallback updates immediately.
- Added `CONFIGURE_DEPENDS` to the project source/header/resource globs so the existing build tree picks up newly added source files like `optimization_anchor.cpp` during `cmake --build`.

Validation: `nix develop -c cmake --build build/generic-llvm-ninja --config Debug` now completes successfully locally.

aroschli:
Pushed one more follow-up in `7511c572`:

- Preserves the existing Z height while dragging seam anchors in both the part view and G-code view, instead of snapping the marker back to the bed plane during X/Y drag.

Validation: `nix develop -c cmake --build build/generic-llvm-ninja --config Debug` still passes locally.

PR #210, opened 2026-07-15
Spiral Contour Improvements

aroschli:
## Summary

This PR improves spiral contour stitching for perimeters and insets.

- Adds width-aware linking for adaptive spiral contour transitions.
- Keeps single-contour spiral paths closed instead of cutting their final segment short.
- Uses single-segment contour-to-contour links and avoids point-order segment breaking during spiral transitions.
- Applies the corrected open-loop tip-wipe handling to both perimeters and insets.

## Root Cause

Spiral contour linking was treating all loops like stitched open paths, even when only one contour existed, and follow-up loop selection could split or rotate paths in a way that introduced extra transition segments. Adaptive contours also needed to pass their actual bead widths into the spiral linker so transition stop points matched the generated contour widths.

## Validation

- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja --target ornlslicer_obj`

### DIFF
--- a/include/geometry/spiral_path.h
+++ b/include/geometry/spiral_path.h
@@ -194,12 +194,23 @@ inline Distance transitionDistance(Distance current_width, Distance next_width, 
 }
 
 inline Point prepareConnector(const Polyline& current_loop, Polyline& next_loop, Distance stop_distance) {
-    Point connector_start = stopPointOnClosingSegment(current_loop, stop_distance);
+    const Point rough_connector_start = stopPointOnClosingSegment(current_loop, stop_distance);
+    const bool smooth_closing_segment = hasSmoothClosingSegment(current_loop);
     if (next_loop.size() >= 3) {
-        rotateToClosestForwardExistingPoint(next_loop, connector_start, current_loop.back(), current_loop.front());
+        if (smooth_closing_segment) {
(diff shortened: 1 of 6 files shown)


aroschli:
Applied the review follow-up in `b249e982`.

- Adaptive spiral perimeter/inset paths now route loop selection back through `PolylineOrderOptimizer`, so configured path-order behavior such as `NextClosest`, `CustomPoint`, `InsideOut`, etc. is preserved.
- The adaptive bead widths are still carried into `SpiralPath::linkClosedPolylines` for width-aware transition distances.
- Removed the manual area/distance loop-ordering helpers that were bypassing the configured optimizer.

Validation:

- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja --target ornlslicer_obj`

aroschli:
Pushed a follow-up fix in `80ee1b82` for the hollow-square spiral perimeter intersections.

What changed:
- Smooth closing segments continue using the cylinder-safe existing-point rotation path.
- Sharp/cornered contours can now rotate the next loop to a closest point along an edge, so the transition does not have to snap across a square corner.
- For cornered contours, the connector start is projected onto the current loop's closing segment to keep the transition along the intended edge.

Validated locally with:
- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja --target ornlslicer_obj`

PR #213, opened 2026-07-15
Fix Skeleton Prestarts

aroschli:
## Summary
- Wire the existing `skeleton_prestart*` profile settings into skeleton path modifier generation.
- Add a prestart path modifier helper that prepends a same-vector open-loop segment and marks it as `PRESTART`.
- Preserve segment width, height, acceleration, region, material, and width/height area modifier behavior for the generated prestart segment.

## Root Cause
The skeleton prestart settings were present in the settings YAML/master config and `PathModifiers::kPrestart` already existed, but the skeleton region code never read those profile settings or generated a prestart segment.

## Validation
- `nix develop --command ninja -C build/generic-llvm-ninja ornlslicer`


aroschli:
Review follow-up complete.

I found one low-risk metadata issue while reviewing the prestart helper: the generated prestart segment copied width/height and most segment settings from the first skeleton segment, but it did not preserve `SS::kAdapted`. Adaptive skeleton output uses that flag downstream for G-code comments and reload/visualization, and the other cloned modifier segments already copy it.

I pushed the fix in `a62ed213` by adding the adapted-width flag copy in `GeneratePrestart`. The PR is now rebased onto current `develop` and GitHub reports it mergeable.

PR #216, opened 2026-07-19
Add sharp corner extension

aroschli:
## Summary

- Adds a Special Modes setting group for sharp corner extension, including enable, angle threshold, and extension distance controls.
- Adds `PathModifierGenerator::GenerateSharpCornerExtension` to detect sharp printing-line junctions and move the shared point outward along the local corner bisector.
- Applies the modifier early in the standard extrusion region modifier flow for perimeter, inset, infill, skin, skeleton, support, raft, brim, and skirt paths.

## Validation

The build completed successfully and linked `Debug/ornlslicer`.


liamnwhite1 (approved):
Fixed the implementation. Should be good to merge now. One thing to note for the future is we may want to visualize the extension segments differently from the path they belong to. Haven't decided on this yet.

PR #217, opened 2026-07-26
Fix skeleton compute crashes

aroschli:
## Summary
Fixes the compute-time skeleton crash reproduced with `New-Impact-Limiter-Bottom.s2p` and avoids a factorial post-compute path in `Next Farthest` island ordering.

## Changes
- Clear skeleton graph/output state for each compute.
- Avoid stale Boost graph edge descriptors by copying endpoints before graph mutation.
- Make `Next Farthest` select greedily instead of invoking longest brute-force TSP.
- Clear cached TSP results when optimizer inputs change.

## Validation
- Built `ornlslicer` with CMake/Ninja in the Nix dev shell.
- Sliced the provided project successfully through compute, postprocess, G-code generation, parsing, and visualization.
- Generated `/tmp/ornl-slice-repro/New-Impact-Limiter-Bottom.gcode`.

### DIFF
--- a/include/optimizers/island_order_optimizer.h
+++ b/include/optimizers/island_order_optimizer.h
@@ -112,6 +112,12 @@ class IslandBaseOrderOptimizer {
      */
     int computeNextClosest();
 
+    /*! \brief Computes the island order with following logic:
+     *          Each time move to the farthest polygon from current position
+     *  \return Index of next island
+     */
+    int computeNextFarthest();
+
     /*! \brief The index of the island that has the nearest end point from a start point
      *  \return Index of next island
      *  \param start_point Starting point to consider for calculation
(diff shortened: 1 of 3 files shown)


aroschli:
Applied the follow-up fix in `8fe52161`.

`Skeleton::extractPath` now stores the exact `SkeletonEdge` descriptors while copying path points, then removes those descriptors after graph reads are done. This preserves the original stale-descriptor crash fix while avoiding endpoint-pair removal, which could delete parallel skeleton edges in the `listS` graph.

Validation:
- `clang-format` on `src/step/layer/regions/skeleton.cpp`
- `git diff --check`
- PCH-free `clang++ -fsyntax-only` check for `src/step/layer/regions/skeleton.cpp` using the configured include roots

PR #218, opened 2026-07-28
Improve Skeleton Pruning

aroschli:
## Summary

Adds an optional Vanishing Angle pruning mode for skeleton generation. The new mode prunes low-significance skeleton graph branch trees after Voronoi graph generation while preserving the existing default behavior when pruning is disabled.

## Changes

- Added skeleton pruning settings for `NONE` and `VANISHING_ANGLE` modes.
- Added a tunable Vanishing Angle threshold with a default around 50 degrees.
- Implemented a graph pruning pass before cycle/simple-path extraction.
- Kept existing bead-width, adaptive-width, and path filtering behavior intact.

## Impact

Existing profiles keep the current skeleton output because the new pruning mode defaults to `NONE`. Users can opt into Vanishing Angle pruning to reduce noisy skeleton branches while preserving topology-sensitive structure.

## Validation

- `git diff --check`

### DIFF
--- a/include/step/layer/regions/skeleton.h
+++ b/include/step/layer/regions/skeleton.h
@@ -78,6 +78,9 @@ class Skeleton : public RegionBase {
     //! \brief Generates a graph representation of skeleton geometry for cleaning
     void generateSkeletonGraph();
 
+    //! \brief Prunes low-significance skeleton graph branches using the Vanishing Angle measure
+    void pruneSkeletonGraphByVanishingAngle();
+
     //! \brief Cleans graph of skeleton geometry according to ClipperLib's CleanPolygon function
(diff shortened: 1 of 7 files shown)


aroschli:
Follow-up fix pushed in `feb68621`.

What changed:
- Skips zero-length boundary edges before calling `nearestPointOnSegment()` so duplicate polygon vertices cannot feed a divide-by-zero projection into the Vanishing Angle score.
- Ignores non-finite nearest-boundary distances and rejects non-finite edge/candidate angle scores before selecting a pruning candidate.
- Keeps pruning disabled for invalid/non-finite threshold values.

Validation:
- `git diff --check`
- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` (no tests registered in this build tree)

PR #221, opened 2026-07-29
Error Checking for Settings Values

aroschli:
## Summary

Adds warning-only validation for settings values that can become internally inconsistent across printer, material, and profile configuration.

## Changes

- Refreshes dynamic setting warnings when related settings change, profiles switch, or rows reload.
- Warns when speed, extruder RPM, printer dimension, bead geometry, infill path length, lift/travel, fan speed, or minimum layer time values violate related configured limits.
- Reuses the existing warning styling and tab warning counts so invalid values are visible without blocking edits.

## Validation

- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` (no registered tests found)
- `git diff --check`

### DIFF
--- a/include/widgets/settings/double_spin_subtypes/setting_speed_spin_box.h
+++ b/include/widgets/settings/double_spin_subtypes/setting_speed_spin_box.h
@@ -3,6 +3,7 @@
 #include <qgridlayout.h>
 #include <qobject.h>
 #include <qsharedpointer.h>
+#include <qstring.h>
 #include <qtmetamacros.h>
 #include <qvariant.h>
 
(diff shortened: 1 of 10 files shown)


aroschli (commented):
Addressed the warning-refresh performance concern in `fe8b7cf0`.

What changed:
- Added a scoped guard so bulk row reloads do not trigger a full dynamic-warning refresh once per emitted row `modified` signal.
- Let `enableDependRows()` remain the single full dependency/dynamic-warning pass after profile/base reload paths, removing redundant follow-up refreshes.
- Moved unit-change reload handling to `SettingBar` so unit changes reload all panes under the guard and then run one final dynamic-warning refresh.

Validation:
- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` (no registered tests found)
- `git diff --check`

aroschli:
Followed up on the review findings in `14742ffa`:

- Dynamic warnings now evaluate each selected settings base when checking peer settings, so multi-selected ranges no longer depend on whichever range happens to be index `0`.
- Dynamic refreshes now apply warning styling/tooltips quietly instead of showing transient `QToolTip` popups for unrelated rows on every edit.

Validation:

- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` (no registered tests found)
- `git diff --check`

PR #223, opened 2026-07-30
Fix for Infill Links that Overlap with Contours

aroschli:
## Summary

Adds a default-off infill setting, `infill_avoid_link_overlap`, that prevents printed infill linking segments from intruding beyond the configured contour overlap allowance.

## Root Cause

Short infill linking moves were being treated like normal infill extrusion links based primarily on centerline intersection checks. For some geometries, the link centerline could pass the existing checks while the printed bead footprint still overlapped contour space.

## Changes

- Adds the `Avoid Infill Link Contour Overlap` profile setting, defaulting to `false` to preserve existing behavior.
- Passes overlap-aware link-check parameters from `Infill` into `PolylineOrderOptimizer`.
- Rejects candidate infill links when an overlap-adjusted bead core footprint exits the overlap-expanded infill geometry, causing those unsafe links to remain travel moves.
- Regenerates `resources/configs/master.conf` from the settings YAML.

## Validation

- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug`
- `git diff --check`
- Headless sliced the provided infill-overlap sample with the setting enabled and confirmed legal links are preserved while contour-overlapping links are converted to travels.

### DIFF
--- a/include/optimizers/polyline_order_optimizer.h
+++ b/include/optimizers/polyline_order_optimizer.h
@@ -51,13 +51,13 @@ class PolylineOrderOptimizer {
     /// @param minDistance: minimum distance of resulting polyline
     /// @param minTravelDistance: minimum distance for travel to be used instead of link
     /// @param enable_partitioned_linking: whether to enable partitioned linking of line infill
-    /// @param avoid_link_overlap: whether infill print-link footprints must stay inside link_overlap_geometry
-    /// @param link_footprint_width: bead footprint width to use when checking infill print-link footprints
+    /// @param avoid_link_overlap: whether infill print-link cores must stay inside link_overlap_geometry
+    /// @param link_core_width: overlap-adjusted bead core width to use when checking infill print links
     /// @param link_overlap_geometry: overlap-expanded geometry that linked infill print-segment cores must remain
     /// inside
     void setInfillParameters(InfillPatterns infillPattern, PolygonList border_geometry, Distance minInfillPathDistance,
                              Distance minTravelDistance, bool enable_partitioned_linking = false,
-                             bool avoid_link_overlap = false, Distance link_footprint_width = 0,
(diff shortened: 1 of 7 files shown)


aroschli:
Pushed follow-up commit `37bb8749` to address the overlap allowance issue found in review.

What changed:
- The infill link overlap check now uses the full infill bead width against `m_geometry_copy`, which is already expanded by `infill_overlap_distance`.
- Renamed the optimizer parameter/member from overlap width to footprint width so the code reflects the actual geometry being checked.
- Updated the setting tooltip from bead core to bead footprint.

Validation:
- `git diff --check`
- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug`

aroschli:
Pushed follow-up commit `5f951535` for the regression where legal infill links on the overlap boundary were being converted to travels.

The previous fix intentionally used the full bead footprint, but the containment check was too strict for links that exactly touch the overlap-expanded boundary. Clipper can return tiny sliver differences in that case, so the optimizer now relaxes the allowed overlap geometry by one internal unit/1 micron before checking the full footprint. This keeps the full-bead guard while preserving legal boundary-touching links.

Validation:
- `git diff --check`
- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug`

aroschli:
Pushed correction commit `764ed8d7` after the full-footprint follow-up proved too strict in preview.

The link-overlap check now uses the overlap-adjusted bead core width again: `bead_width - 2 * infill_overlap_distance`, checked against the overlap-expanded infill geometry. That restores the intended legal links while retaining the tiny 1 micron expansion from `5f951535` only as protection against Clipper slivers on exact boundary contact.

Validation:
- `git diff --check`
- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug`

PR #224, opened 2026-07-30
Add Layer Time Estimate to G-Code

### LINKED ISSUE #64: Layer times in output file
The layer time information is currently only displayed in the UI.
It could be combined with the gcode output at the `;BEGINNING LAYER: n` lines.

aroschli:
## Summary
- add a Printer Settings -> G-Code checkbox for layer time comments
- annotate each generated `BEGINNING LAYER` marker with an adjusted layer time estimate
- reuse the existing G-code parser layer-time calculation so the emitted value matches adjusted layer-time behavior

Closes #64

## Validation
- `git diff --check`

### DIFF
--- a/include/gcode/parsers/common_parser.h
+++ b/include/gcode/parsers/common_parser.h
@@ -559,6 +559,9 @@ class CommonParser : public ParserBase {
     // List of all calculated times/volumes for layers as well as total distance
     QList<Time> m_layer_times;
 
+    //! \brief Additional layer time inserted by dwell-based minimum layer time enforcement.
+    QList<Time> m_layer_dwell_adjustments;
+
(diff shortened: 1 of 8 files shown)


aroschli:
Applied the follow-up fixes in `49f33cf5`:

- Layer time comments now map each `BEGINNING LAYER: n` marker to the correct adjusted-time entry even when the parser does not have a pre-layer bucket, covering the `Haas-Metric-No-Comments`/no `LAYER COUNT` case.
- Dwell-based minimum layer time adjustments are now tracked separately and included in `getAdjustedLayerTimes()`, so emitted layer-time comments reflect dwell insertion without changing the raw layer-time list.

Validation:
- `git diff --check origin/develop...HEAD`

PR #225, opened 2026-07-31
Arc Visualization Color Preference & Fix Model Improvements

### LINKED ISSUE #15: Arc Color Preference
## Expected Behavior
Self-Explanatory.

## Actual Behavior
Arcs can be tricky to visualize

## Possible Solution
Implement a color preference for arc segments so that arcs are easier to visualize with respect to their linear neighbors. For example separate color preferences for perimeter and perimeter arc paths. Should also apply to insets.

Self-Explanatory.

## Specifications
N/A

aroschli:
## Summary
- Adds separate visualization color preferences for perimeter arcs and inset arcs.
- Applies those colors when loaded G-code commands are G2/G3 arc moves in perimeter or inset regions.
- Keeps modifier colors, such as tip wipe and slow down, taking precedence over arc-specific colors.

## Why
Issue #15 notes that arcs are hard to distinguish from neighboring linear segments in the G-code preview. The new preferences make arc geometry easier to identify without replacing the existing region-color scheme.

## User Impact
Users can customize `PerimeterArc` and `InsetArc` in the existing Visualization Colors preferences. The defaults stay close to the existing perimeter and inset blues while remaining visually distinct.

## Validation
- `nix develop --command cmake --build --preset debug-generic-llvm-ninja`

Closes #15

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,8 @@ project(
     HOMEPAGE_URL "https://github.com/ORNLSlicer/ORNLSlicer"
 )
 
+include(CTest)
+
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
@@ -223,7 +225,7 @@ if(WIN32)
     target_link_libraries(${PROJECT_NAME}_cli ${PROJECT_NAME}_obj)
 endif()
 
-option(ORNLSLICER_BUILD_TESTS "Build ORNLSlicer test executables" ON)
+option(ORNLSLICER_BUILD_TESTS "Build ORNLSlicer test executables" ${BUILD_TESTING})
 if(ORNLSLICER_BUILD_TESTS)
     enable_testing()
     add_executable(${PROJECT_NAME}_mesh_repair_tests tests/mesh_repair_tests.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,6 +137,7 @@ set(ORNLSLICER_LIBRARY_DIRS ${RESULT_LIBRARY_DIRS})
 file(GLOB_RECURSE SOURCES CONFIGURE_DEPENDS "src/**.cpp")
 file(GLOB_RECURSE HEADERS CONFIGURE_DEPENDS "include/**.h")
 file(GLOB_RECURSE RESOURCES CONFIGURE_DEPENDS "resources/**.qrc")
+list(REMOVE_ITEM SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp")
 
 add_library(${PROJECT_NAME}_obj OBJECT)
 target_include_directories(
@@ -208,6 +209,8 @@ if(ORNLSLICER_ENABLE_UNITY_BUILD)
 endif()
 
 add_executable(${PROJECT_NAME})
+target_sources(${PROJECT_NAME} PRIVATE src/main.cpp)
+target_include_directories(${PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/generated/include)
 target_link_libraries(${PROJECT_NAME} ${PROJECT_NAME}_obj)
 
 if(WIN32)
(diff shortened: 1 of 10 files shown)


aroschli:
Updated the arc visualization defaults so they stay closer to the existing perimeter/inset blues:

- `InsetArc`: `QColor(0, 184, 255, 255)`
- `PerimeterArc`: `QColor(32, 64, 255, 255)`

Validation: `nix develop --command cmake --build --preset debug-generic-llvm-ninja` passed.

aroschli:
Pushed follow-up commit `516f74cd` for the STL load crash path.

What changed:
- `ClosedMesh::CleanPolyhedron` no longer mutates the polyhedron while iterating the original `halfedges(polyhedron)` range.
- Hole filling now repairs one current border halfedge at a time, then restarts from a fresh border lookup after CGAL mutates the mesh.
- Added bounded retry protection and exception handling so repair failure returns `false` and the loader can continue with the unrepaired open mesh instead of falling into the crashy repair path.

Validation:
- `nix develop --command cmake --build --preset debug-generic-llvm-ninja` passed.
- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` ran, but no tests were registered.
- I did not rerun ORNLSlicer against `2DIceCreamCone.STL` because that repro path previously faulted hard enough to drop the WSL connection.

Also tracked down the apparent arc color regression: the source defaults are still the blue arc colors from the prior commit (`PerimeterArc=#2040ff`, `InsetArc=#00b8ff`), but the active local preferences had persisted the old magenta/yellow values. I updated the local preferences file locally; no source change was needed for that part.

aroschli:
Pushed follow-up commit `9acb52bb` for the arc color issue.

Root cause: the checked-in arc defaults were already blue, but existing `app.preferences` files could still contain stale arc entries (`PerimeterArc=#ff00cc`, `InsetArc=#ffb300`). Since persisted preferences override defaults, changing `VisualizationColorsDefaults()` alone was not enough for anyone who had already run a build with those values saved.

What changed:
- Added a targeted visualization color migration in `PreferencesManager::setDefaultVisualizationColors()`.
- If persisted `PerimeterArc` / `InsetArc` match the stale magenta/yellow values, or the earlier PR defaults, they are replaced with the current blue defaults (`#2040ff` / `#00b8ff`).
- Other visualization colors, including real user customizations outside those stale arc values, are left alone.

Validation:
- `nix develop --command cmake --build --preset debug-generic-llvm-ninja` passed.
- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` ran, but no tests were registered.

I also reset the local app preferences file again so the next local run should immediately pick up `PerimeterArc=#2040ff` and `InsetArc=#00b8ff`.

aroschli:
Pushed follow-up commit `9254d4f5` to further constrain the fix-model path for open/STL article geometry.

Additional diagnosis on `2DIceCreamCone.STL`: the file is very small (`60` triangles) but has `36` boundary edges, so it is not a good candidate for automatic hole reconstruction. The prior change avoided mutating stale halfedge iterators, but the repair path could still spend time in CGAL trying to infer/repair an intentionally open model.

What changed:
- Hole filling now uses bounded plain `triangulate_hole()` instead of `triangulate_refine_and_fair_hole()`, avoiding refinement/fairing and sparse-solver work during import repair.
- Large/open boundary cases are skipped before hole filling, causing repair to return `false` and the loader to fall back to `OpenMesh` quickly.
- Self-intersection auto-refinement now only runs after the mesh has been confirmed closed, so open STLs do not enter that heavier repair step unnecessarily.

Validation:
- `nix develop --command cmake --build --preset debug-generic-llvm-ninja` passed.
- `nix develop --command ctest --test-dir build/generic-llvm-ninja --output-on-failure` ran, but no tests were registered.
- Lightweight STL topology check for `2DIceCreamCone.STL`: `faces=60 boundary_edges=36 repair_limit=32 would_skip=True`.

I still did not launch the full slicer against this STL from Codex because that repro path has previously been severe enough to drop the WSL connection.

aroschli:
Manual follow-up validation: the constrained fix-model change in `9254d4f5` resolved the `2DIceCreamCone.STL` load issue with fix model enabled.

No additional code changes from this verification step.

aroschli:
Pushed follow-up commit `b6c86e52` with the fix-model hardening items we discussed.

What changed:
- Added `ClosedMesh::RepairResult` plus `CleanPolyhedronWithStatus()` so repair callers can distinguish success from skip/failure reasons instead of only getting a `bool`.
- Kept `CleanPolyhedron()` as a compatibility wrapper for existing callers.
- Added a self-intersection preflight with `does_self_intersect()` so the expensive CGAL auto-refinement path only runs when the mesh is closed and actually reports self-intersections.
- Updated STL/Assimp and STEP import repair paths to run repair on a copy and log a clearer reason before importing unrepaired geometry.
- Added a small CTest regression target, `mesh_repair_tests`, covering large-boundary open meshes being skipped and ordinary closed meshes still succeeding.
- Adjusted CMake so `src/main.cpp` is compiled into the app targets instead of the shared object library, allowing tests to link project objects without a duplicate `main()`.

Validation:
- `nix develop --command cmake --build --preset debug-generic-llvm-ninja` passed.
- `nix develop --command ctest --test-dir build/generic-llvm-ninja -C Debug --output-on-failure` passed: `1/1 Test #1: mesh_repair_tests`.

I did not add a Conservative/Aggressive UI repair mode in this PR; that seems better as a separate settings/UI follow-up after this safer default behavior lands.

aroschli:
Pushed follow-up commit `3e3a97a8` applying the review fixes.

What changed:
- Made the stale arc visualization color migration versioned/one-time via a persisted `visualization_color_migration_version`, so users can intentionally choose those old colors after the migration has run.
- Changed arc modifier precedence to use comment semantics instead of `QColor` equality, so custom color collisions cannot prevent `PerimeterArc` / `InsetArc` colors from applying.

Validation:
- `cmake --build --preset debug-generic-llvm-ninja` passed.
- `ctest --test-dir build/generic-llvm-ninja -C Debug --output-on-failure` passed: `1/1 Test #1: mesh_repair_tests`.

aroschli:
Follow-up on the failed Nix CI run from `3e3a97a`:

- The failed GitHub check was the Windows `Generate Installer` step. The public Actions page did not expose the NSIS log without an authenticated GitHub session.
- I reproduced the Windows package locally with constrained parallelism and then ran the same NSIS command from CI. The local installer generation succeeded and produced a 112 MB installer.
- I noticed the Nix/CMake package path was still compiling `ornlslicer_mesh_repair_tests` even though Nix passes `BUILD_TESTING=OFF`, so I pushed `cfb38cea` to wire `ORNLSLICER_BUILD_TESTS` to CMake's standard `BUILD_TESTING` option.
- Validation: local debug build passed, `ctest` passed, `BUILD_TESTING=OFF` configure no longer exposes the mesh repair test target, and the fresh GitHub Nix CI run for `cfb38cea` is green across ubuntu check, macOS check, linux, and windows.

PR #227, opened 2026-08-01
Fix consecutive point ordering

### LINKED ISSUE #10: Consecutive Start Stop Increments on Index Point rather than Absolute XY value
## Expected Behavior
Start/stop location will be at least as far apart as set in "Consecutive Distance Threshold" setting.

## Actual Behavior
Sometimes start stops are exactly on top of each other, or only spaced apart by less than the threshold distance. For example, in the attached project, layers 23-24, and layers 25-26, and layers 39-40, the start stop is right on top of each other. Layers 26 to 34 the start stop is staggered, but only by maybe 1/4" when the threshold is set to 1". This is true even though the geometry only changes slightly from layer to layer.

## Possible Solution
Modify the algorithm to work on XY coordinates in the gcode rather than an "N" index point within the layer. Or, if this is computationally intensive, enable a user checkbox to do this and the user would have to understand that it will take longer to slice then.

Open the project file and slice. View the gcode and scroll through the layers.
image

[Consec.Start.Stop.Single.STL.Increment.Not.Enough2.zip](https://github.com/ORNLSlicer/Slicer-2/files/13756480/Consec.Start.Stop.Single.STL.Increment.Not.Enough2.zip)

aroschli:
## Summary
- Fix consecutive point ordering to use the previous layer's actual physical start point instead of only the layer-derived vertex index.
- Allow closed contours to split at the exact XY threshold distance when the target falls along a segment.
- Add a focused point-order optimizer regression test and register it with CTest.

## Root Cause
The consecutive start/stop optimizer selected the next seam from `layer_number - 2` and walked along the current polyline. That made the configured threshold depend on contour index position, so similar layers could still start too close together in XY space.

## Validation
- `ctest --test-dir build/generic-llvm-ninja -C Debug -R 'mesh_repair_tests|point_order_optimizer_tests' --output-on-failure`

Fixes #10

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,6 +238,17 @@ if(ORNLSLICER_BUILD_TESTS)
     )
     target_link_libraries(${PROJECT_NAME}_mesh_repair_tests PRIVATE ${PROJECT_NAME}_obj)
     add_test(NAME mesh_repair_tests COMMAND ${PROJECT_NAME}_mesh_repair_tests)
+
+    add_executable(${PROJECT_NAME}_point_order_optimizer_tests tests/point_order_optimizer_tests.cpp)
+    target_include_directories(
+        ${PROJECT_NAME}_point_order_optimizer_tests
+        PRIVATE
+            ${CMAKE_CURRENT_LIST_DIR}/include
+            ${CMAKE_CURRENT_BINARY_DIR}/generated/include
+            ${ORNLSLICER_INCLUDE_DIRS}
+    )
+    target_link_libraries(${PROJECT_NAME}_point_order_optimizer_tests PRIVATE ${PROJECT_NAME}_obj)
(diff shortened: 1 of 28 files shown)


aroschli:
Applied the review follow-ups in `f76fb111`:

- Preserved the consecutive split selection when local randomness has no eligible nearby vertices, avoiding an empty-candidate crash/assert path.
- Added a regression case for that split-plus-randomness edge case in `point_order_optimizer_tests`.
- Added skirt regions with generated paths to `previousRegions` so multi-layer skirts can provide the prior start point for consecutive ordering.

Validation:
- `git diff --check`
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_point_order_optimizer_tests ornlslicer_mesh_repair_tests`
- `nix develop -c ctest --test-dir build/generic-llvm-ninja -C Debug -R 'mesh_repair_tests|point_order_optimizer_tests' --output-on-failure`

PR #229, opened 2026-08-04
Fix installer slicing crash after compute

aroschli:
## Summary

Fixes a crash seen when slicing a large test project in the installed build path. The local/debug executable now completes compute, planar post-processing, G-code generation, parsing, and visualization for that project.

## Root Cause

The deterministic crash reproduced during compute, not after post-processing started. `Skeleton::extractPath()` could collect the same undirected Boost graph edge more than once, then remove the original edge descriptor repeatedly. With the list-backed Boost edge container, the second stale-descriptor removal could corrupt/unhook the edge node and segfault.

## What Changed

- Deduplicate skeleton edge removals by endpoint pair and re-resolve each edge before removing it.
- Add visible runtime/build diagnostics and compute-to-postprocess phase breadcrumbs for installed-build triage.
- Emit `Post-Process: 0` at planar post-process entry and log per-layer island/region/path counts.
- Harden planar post-processing and path/island/polyline ordering against empty or null geometry.
- Stop worker `StepThread` instances before deletion.
- Add focused empty-input optimizer tests and an optional local `new-impact.s2p` regression CTest that asserts non-empty G-code output.

## Validation

- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer_optimizer_empty_input_tests ornlslicer -j 4`
- Result: `3/3` tests passed, including `new_impact_project_regression`.
- Regression output: generated non-empty `new-impact.gcode` (~6.5 MB).
- `build/generic-llvm-ninja/Debug/ornlslicer --version` now prints version, git revision/dirty state, build configuration, package type, and Qt runtime/build versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ set(CMAKE_AUTOUIC OFF)
 set(CMAKE_AUTORCC ON)
 
 include(build_info)
+set(ORNLSLICER_PACKAGE_TYPE "local" CACHE STRING "Package/distribution type reported in runtime diagnostics.")
 FetchBuildInfo()
 set(ORNLSLICER_BUILD_VERSION "${RESULT_VAR}")
 configure_file(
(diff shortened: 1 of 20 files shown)


aroschli:
Follow-up changes pushed in `49469d0c` to address the review findings:

- Fixed farthest path/polyline selection so zero-distance candidates are still selected and consumed, avoiding no-progress loops in callers that iterate on remaining count.
- Updated `GlobalLayer` spiral-lift handling to mutate the stored final path by reference instead of appending lift segments to a local copy while advancing `current_location`.
- Updated brim sequencing in `Layer` and `GlobalLayer` to filter invalid brim parents first and fall back to top-level child islands when no valid brim geometry remains.
- Extended `optimizer_empty_input_tests` with zero-distance farthest-order coverage for both path and polyline optimizers.

Validation after the follow-up:

- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer_optimizer_empty_input_tests -j 4`
- Result: `3/3` tests passed, including `new_impact_project_regression`.

PR #230, opened 2026-08-05
Add G-Code to S2C importer

### LINKED ISSUE #228: G-Code to S2C
## Summary
Create a settings s2c file from an existing g-code file by reading the raw setting values from the footer.

## Problem or opportunity
Old g-code files may not have a record of the settings used to create them.

## Proposed capability
Add a new window, under the settings dropdown on the menu bar, to create the s2c file. The window should ask you to select the g-code file and give a name for the new settings file. A checkbox is needed for using the default values for settings that can't be found, or for prompting the user to input a value for each setting that doesn't exist. Upon completion the settings file should be loaded into the Slicer and be ready for use.
Error checking should persist for invalid values, missing values, and values that would fail the existing settings error checking.

aroschli:
## Summary
- Add a Settings > G-Code to S2C workflow for creating `.s2c` files from G-code settings footers.
- Add a reusable `GcodeSettingsImporter` that parses footer values, fills missing settings from defaults or prompts, validates active settings, and formats a current-version settings file.
- Add importer tests for semicolon and parenthesized footer formats.

## Impact
Users can recover a complete settings template from old G-code files that contain ORNLSlicer settings footers, then immediately load that template into the Slicer.

Closes #228.

## Validation
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_gcode_settings_importer_tests -j2`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,17 @@ if(ORNLSLICER_BUILD_TESTS)
     target_link_libraries(${PROJECT_NAME}_optimizer_empty_input_tests PRIVATE ${PROJECT_NAME}_obj)
     add_test(NAME optimizer_empty_input_tests COMMAND ${PROJECT_NAME}_optimizer_empty_input_tests)
 
+    add_executable(${PROJECT_NAME}_gcode_settings_importer_tests tests/gcode_settings_importer_tests.cpp)
+    target_include_directories(
+        ${PROJECT_NAME}_gcode_settings_importer_tests
+        PRIVATE
+            ${CMAKE_CURRENT_LIST_DIR}/include
+            ${CMAKE_CURRENT_BINARY_DIR}/generated/include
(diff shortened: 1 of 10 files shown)


aroschli:
Applied the review fixes in `cdd8dd9f`:

- Migrated known legacy footer setting keys before importer validation so old G-code footers with renamed slicing settings are recovered instead of treated as unknown/defaulted.
- Changed footer parsing to stream line-by-line instead of reading/splitting the entire G-code file in memory.
- Made missing-setting prompt cancellation stop the import immediately.
- Added a dialog guard to prevent the output `.s2c` path from overwriting the selected G-code file.
- Added importer coverage for legacy footer keys and prompt cancellation.

Validation:
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_gcode_settings_importer_tests -j2`

All 4 configured Debug tests passed.

PR #232, opened 2026-08-06
Add dynamic optimization point visualization

aroschli:
## Summary

- Adds an effective selected-settings signal from the settings bar for optimization point visualization.
- Merges global, inherited part, and selected local override settings for the first selected target.
- Wires the part and G-code views to draw optimization points from that selected-settings snapshot without replacing each view's primary settings base.

## Why

When editing part or layer-range seam attractor settings, the visualization continued to show the global attractor. This makes the preview misleading for local optimization settings. The new flow lets the preview reflect the active settings editor target while still falling back to global settings when nothing local is selected.

## Validation


aroschli:
Applied the review fixes in `2e2185a9`.

- Paired optimization point dragging now edits the currently selected local settings bases when present, falling back to Global only when no local target is selected.
- Drag completion and undo/redo restore paths now emit the selected visualization settings snapshot after `settingModified`, so the preview refreshes from the effective selected settings instead of stale/global settings.

Validation:
- `git diff --check`

aroschli:
Follow-up fix pushed in e2bb55fe: the Seam Attractor Vector setting now only shows for Point Order Optimization = Custom Location, and runtime projection ignores saved seam vectors unless that strategy is selected. Added optimization_anchor_tests coverage and confirmed the focused test plus local ornlslicer build pass. New Nix CI run queued: https://github.com/ORNLSlicer/ORNLSlicer/actions/runs/31196340241

aroschli:
Follow-up pushed in b4df8718: Seam Attractor Vector is now visible and active for both Point Order Optimization custom modes: Custom Location and Custom Farthest Location. The runtime guard now uses the shared custom-point-location helper, and optimization_anchor_tests now asserts that Custom Farthest Location uses the seam attractor vector. Local focused test and ornlslicer build both pass. New Nix CI run queued: https://github.com/ORNLSlicer/ORNLSlicer/actions/runs/31199944922

aroschli:
Pushed follow-up fix in `3e45a428`.

- Restored seam attractor vector support for custom island/path anchors while keeping the point-order guard for non-custom point ordering.
- Aligned the visualization guide with the same active custom-anchor rules so it no longer depends only on point order state.
- Cleaned up dragged local point edits so selected part/range overrides are removed again when the dragged value matches the inherited value.
- Expanded `optimization_anchor_tests` to cover island/path custom projection independent of point order.

Validation:
- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja --target ornlslicer_optimization_anchor_tests -j2`
- `nix develop --command ctest --test-dir build/generic-llvm-ninja -C Debug -R optimization_anchor_tests --output-on-failure`

PR #234, opened 2026-08-07
Fix Arc Specialties startup approach positioning

liamnwhite1:
## Summary

Fixes Arc Specialties startup positioning so the first approach is emitted in a TRAFO-off world frame before work-object kinematics are enabled. This also adds Arc Specialties G-code documentation and tidies the related issue/PR templates.

## Related issues

- No related issues.

## Details

The Arc Specialties writer now starts with `#TRAFO OFF`, defers the controller kinematics block until the first travel target is known, and emits an `INITIAL WORLD APPROACH` move before enabling work-object kinematics. The first world approach uses a safe Z value based on the build maximum Z plus a 100 mm buffer, and cylindrical output uses the configured cylinder center XY for that approach.

The writer now records the maximum build Z from non-clipping parts, supports inverse G-code frame rotation for first-travel lift alignment, and defers the first layer marker until after startup kinematics are written. Absolute arc center shutdown cleanup is only emitted if startup kinematics actually ran.

Documentation now separates cylindrical slicing workflow from Arc Specialties controller output details. `docs/gcode/arc-specialties.md` documents the dialect, motion fields, AP/CP behavior, arc center modes, startup sequence, travel commands, and parser behavior. The cylindrical slicing page was moved under `docs/slicing/`, and architecture/README links were updated.

The branch also standardizes spacing and wording in the GitHub issue templates and updates the PR template heading from `Testing strategy` to `Validation`.

## Impact

This changes Arc Specialties startup G-code ordering and first-travel positioning. The intended behavior is safer initial motion: the machine approaches above the part in world coordinates before enabling the Arc Specialties work-object kinematics.

Risk level: Medium. The change is isolated to Arc Specialties writer behavior plus shared writer metadata for build maximum Z, but it affects machine startup sequencing and generated controller output. Non-Arc-Specialties writers should be unaffected except for the new optional build-maximum-Z state on `WriterBase`.

## Validation

- Ran `git diff --check origin/develop...HEAD`; passed with no whitespace errors.
- Ran `cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer`; completed successfully with `ninja: no work to do`.


aroschli:
Addressed the review follow-ups in `582c5012`:

- Guarded Arc Specialties short-travel welder-on behavior so it cannot run before the first-travel startup/world-approach sequence.
- Updated the PR template wording to refer to `Validation` instead of the old `Testing strategy` section.

Validation run locally:

- `git diff --check`
- `nix develop --command cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer`

Issue #235, opened 2026-08-07
"As Printed" Model


aroschli:
## Summary
Create a 3D representation of the as built model that shows the 3D printed beads with scalloped edges.

## Problem or opportunity
Many large format 3D printed objects are machined or post-processed after printing. Having a 3D representation of the expected geometry is useful to understand where material will be deposited.

## Proposed capability
Create a 3D representation based on extrusion commands in the g-code. This could be as simple as a 2D profile sweep along the toolpath coordinates, or a more elaborate geometrical extrusion. Libraries like CADquery and CADSmith may be helpful (other options should also be considered), but the preference is to not include a large library that would impact compile times of ORNL Slicer. Perhaps generating a script that could be run using one of the other libraries would be sufficient.

This model should not be generated every time, it should be enabled/disabled via checkbox in the output options or could be an option as part of exporting g-code/project. Open to suggestions for this.

Support structures should not be included in the model.

aroschli:
Closed in PR #238

PR #238, opened 2026-08-08
Add as-printed STL export from G-code

aroschli:
## Summary

Adds an optional as-printed STL companion export generated from parsed G-code bead visualization geometry.

## What changed

- Added an `AsPrintedModelExporter` that writes binary STL by default, with ASCII available for tests/debugging.
- Added an export checkbox to the G-code/project export window for saving `<part>_as_printed.stl`.
- Converts preview-space G-code bead geometry back into millimeter STL coordinates and normalizes output to local origin by default.
- Adds corner blend geometry so connected bead segments fill vertex gaps in exported as-printed models.
- Tracks deposition state on visualization segments so non-deposition/table/safety moves do not expand the exported STL bounds.

## Why

The previous implementation path could leak ORNLSlicer internal distance units and LFAM machine-frame offsets into the STL, producing extremely large geometry that was difficult to locate. This keeps the companion mesh inspectable and aligned to a local model-space artifact.

## Validation

- `nix develop --accept-flake-config --command ctest --test-dir build/generic-llvm-ninja -C Debug -R as_printed_model_exporter_tests --output-on-failure`

Refs #235


aroschli:
Applied the two review fixes in `0efe374a`:

- Cleared and disabled stale as-printed visualization data when a new G-code load starts, and guarded export so the companion STL cannot be written before fresh visualization data is available.
- Kept same-endpoint full-circle arcs exportable by making the degenerate-segment filter line-specific, with a regression test for full-circle arc triangle generation.

Validation:

- `nix develop --accept-flake-config --command cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer_as_printed_model_exporter_tests`
- `nix develop --accept-flake-config --command ctest --test-dir build/generic-llvm-ninja -C Debug -R as_printed_model_exporter_tests --output-on-failure`
- `git diff --check`

Issue #237, opened 2026-08-07
Object Dimensions and Measurements


aroschli:
## Summary
Add the ability to measure an object, from one object to another object, and see the size of an object.

## Problem or opportunity
Object sizes can be hard to interpret in the visualization. 

## Proposed capability
Add a measurement tool, perhaps to the panel on the left side of the visualization, that activates measurement mode. This would allow you to select an edge to see it's length, or select a point/edge/face to measure from one to another.

Add a right click option to objects titled "Info" that would pull up a small window with information such as object name, type (STL/STEP/3MF etc.), number of triangles in the mesh, and dimensions of the object.

aroschli:
Resolved in PR #239

PR #239, opened 2026-08-08
Add object measuring tools

aroschli:
## Summary

Adds object measurement support to the object view and fills out the object info workflow requested in issue #237.

## What changed

- Added a toolbar measurement mode with a monochrome icon that matches the existing part toolbar style.
- Added click-to-measure behavior in the 3D object view: first click starts, mouse movement previews, second click commits a persistent annotation.
- Added explicit clear paths for measurements: Esc, disabling measurement mode, starting a new measurement, or the clear button beside the readout.
- Added a compact measurement readout beside the toolbar and logs completed measurements to the bottom-right Status console.
- Added right-click object Info with file type, mesh role, triangle count, current dimensions, original mesh dimensions, and source path.
- Fixed original mesh dimension reporting for open meshes/imported meshes so Info does not show 0 x 0 x 0 when original vertices have valid bounds.

## Why

Issue #237 requested object dimensions/measurements, including an Info window and a way to measure two picked locations on an object. The follow-up UX changes make the measurement result persistent instead of disappearing as the mouse moves.

## Impact

Users can inspect object metadata from the right-click menu and measure surface-to-surface distances directly in Object View using their configured distance units.

## Validation

- `cmake --build build/generic-llvm-ninja --config Debug -j2`
- `ctest --test-dir build/generic-llvm-ninja -C Debug --output-on-failure`

All 6 tests passed.

### DIFF
--- a/include/graphics/base_view.h
+++ b/include/graphics/base_view.h
@@ -188,6 +188,8 @@ class BaseView : public QOpenGLWidget, public QOpenGLFunctions_3_3_Core {
     virtual void handleWheelForward(QPointF mouse_ndc_pos, float delta);
     //! \brief Respond to a mouse wheel backward movement
     virtual void handleWheelBackward(QPointF mouse_ndc_pos, float delta);
+    //! \brief Gives derived views first chance at non-camera key presses.
+    virtual bool handleKeyPress(QKeyEvent* e);
     //! \brief Applies a pan request from mouse or keyboard navigation.
     //! \param v World-space translation vector to apply.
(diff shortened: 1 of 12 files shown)


aroschli:
Addressed the review finding in `29de12f5`.

What changed:
- Measurement preview/full-clear paths now request a repaint when they actually remove annotation render objects.
- The measurement-object removal helper now reports whether it removed anything, so clear paths avoid unnecessary repaint calls.

Validation:
- `git diff --check develop...HEAD`
- `nix develop -c cmake --build build/generic-llvm-ninja --config Debug -j2`

All 6 tests passed.

PR #245, opened 2026-08-11
Fix Arc Specialties inline G81 arcs

aroschli:
## Summary

- Emit Arc Specialties `G81` inline on every `G02`/`G03` arc when the optional-stop setting is enabled.
- Ignore inline `G81` tokens while parsing Arc Specialties `G2/G3` arcs so generated files can be reloaded without parser errors.
- Update Arc Specialties documentation and setting tooltip text.
- Add a focused regression test for writer output and parser handling.

## Root Cause

The writer kept the old first-deposition guard and emitted `G81` as a standalone line for later arcs. Once `G81` moved inline after `F`, the parser also needed to tolerate the extra `G81` token on arc commands.

## Validation

- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_arc_specialties_parser_tests`
- `nix develop -c ctest --test-dir build/generic-llvm-ninja -C Debug -R arc_specialties_parser_tests --output-on-failure`
- `git diff --check`


liamnwhite1 (approved):
Everything looks good on my end!

PR #261, opened 2026-08-26
Fix flowrate calculator preference units

aroschli:
## Summary
- Update the flowrate calculator labels for bead width, layer height, and gantry speed from application unit preferences.
- Convert existing bead width and layer height entries when the distance unit preference changes.
- Convert calculated gantry speed into the preferred velocity unit while keeping the existing imperial mass-flow calculation intact.

## Verification


liamnwhite1 (approved):
Looks good to me.

PR #262, opened 2026-08-26
Fix JuggerBot bead area multiplier handling

aroschli:
## Summary
- Enable profile region extrusion multiplier settings for JuggerBot pellet width/height mode while keeping RPM controls disabled in that mode.
- Apply the active region extrusion multiplier to JuggerBot width/height bead-area `M3 S...` commands.
- Track scaled bead area so region changes with identical width/height but different multipliers still emit bead-area updates.

## Verification
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_obj`
- `git diff --check`


liamnwhite1 (approved):
Eveything looks good!

PR #268, opened 2026-08-28
Add variable layer height

### LINKED ISSUE #231: Variable Layer Height
## Summary
A function to automatically change the layer height based on perceived geometry changes layer to layer.

## Background
Static layer heights are good for most builds, but certain geometries would benefit from dynamically changing layer height 

## Proposed solution
Preference should always be towards larger layer heights, but swapping to small layer heights (higher resolution) based on changes in the geometry should be added. This must define a standard layer height, and an acceptable minimum layer height. The standard layer height should be used, except for situations where the layer area changes drastically such that going to a small layer height would produce a preferable result.

aroschli:
## Summary
- Add JuggerBot-only variable layer height settings with a target surface error control
- Use a cusp/surface-error estimate from mesh face normals to choose per-layer heights between the configured minimum and standard layer height
- Pass adaptive layer heights through layer/segment settings so JuggerBot width/height bead-area commands use the selected layer height
- Add layer-height/nozzle validation and focused regression coverage

Closes #231

## Verification
- `nix develop -c cmake --build --preset debug-generic-llvm-ninja --target ornlslicer_variable_layer_height_tests`
- `nix develop -c ctest --test-dir build/generic-llvm-ninja -C Debug -R variable_layer_height_tests --output-on-failure`
- `git diff --check`


liamnwhite1:
1. On `src/widgets/layerbar.cpp:1465`:

   The layer count now depends on Variable Layer Height, Minimum Layer Height, Target Surface Error, and machine syntax, but this handler only recalculates for the standard layer height and slicing-axis settings. Enabling adaptive slicing or changing its parameters therefore leaves the LayerBar count and settings-range bounds stale until the part is reselected. Include all adaptive inputs and syntax in this invalidation condition.

2. On `src/widgets/layerbar.cpp:1357`:

   This call uses `computeSliceCount`’s default `include_build_plate_gap=false`, while planar production slicing constructs `Preprocessor` with `include_build_plate_gap=true`. For an elevated part with support enabled, the real slicer creates layers from Z=0 through the gap, but LayerBar counts only from the mesh minimum, producing incorrect layer numbers and settings-range bounds. Pass `true` here or share the production slicing mode explicitly.

liamnwhite1 (approved):
Fixed two issues related to the layer bar not being updated correctly. Otherwise looks good and can be merged!

PR #269, opened 2026-08-28
Define STEP STL resolution preference

### LINKED ISSUE #267: Define STL Resolution from STEP
## Summary
Printing from STEP files is now supported. STL file resolution is automatically determined in the background. Functionality to define STL resolution is needed.

## Proposed solution
Give the user the ability to define high/low resolution. Perhaps as a slider or an input box to define tolerance so that the user can choose how high of a resolution the resultant STL will be.
Functionality should be added to the application preferences.

aroschli:
## Summary
- Add a persisted STEP STL tolerance preference under Preferences > Parts
- Use the preference as the OpenCascade linear deflection when triangulating STEP files
- Add a focused preferences manager regression test for JSON export, round-trip import, and minimum clamping

Closes #267

## Testing
- `direnv exec . cmake --build build/generic-llvm-ninja --target ornlslicer_preferences_manager_tests`
- `direnv exec . ctest --test-dir build/generic-llvm-ninja -C Debug -R preferences_manager_tests --output-on-failure`


liamnwhite1 (approved):
These changes look good to me! Just had to rebase it on develop but once the CI completes this can be merged.

PR #270, opened 2026-08-31
Refactor Show part names in view and Shader

### LINKED ISSUE #12: Corner Ramping Based on Path Type

### LINKED ISSUE #233: Show part names in view broken
## Summary

The `Show part names in view` functionality does not correctly display the names of the parts in the view.

## Steps to reproduce

1. Load a model.
2. Click the `Show part names in view` button located on the controls bar at the top of the view.

## Observed behavior

A white bar appears over the part with no text.

Image

## Expected behavior

A label should appear over the part and should display the name of the part.

## Environment

The issue was observed running the application from the development environment on Ubuntu 24.04.

### LINKED ISSUE #271: text_object Refactor
## Summary
Uses segmenting a font image to create text onto a `QImage` when `DrawText` should probably be used. Background of the text should be transparent. Old comments should be updated & `setText` shouldn't repeat code.

## Background

`Show part names in view feature` will be fixed in https://github.com/ORNLSlicer/ORNLSlicer/pull/270 and could use touchup for easier usability.

Rotating makes the text become upside down at certain angles
Image

jkwookee:
## Summary

Refactors part name display to be a 2d overlay rather than a 3d texture.

Refactors shader lighting calculations and improves barycentric handling

Removes `Text_Object` and `createMeasurementLabel`
## Related issues

- Closes #233 
- Closes #271 

## Details

Produced by creating a part with a name then clicking the option to `Show part names in view`. When a `QImage` was passed to `populateGL` it would act as texture data but since texture data was not sampled in the shader the texture would not render. Fix adds it to the shader. Feature originally worked about 2 years ago but f6ac6ba9f75b8ab85306d23ccd1e235fa6c336ad removed texture sampling within the shader.

Originally a texture was being used on a 3d object for the shader to render and display the text but now is directly drawn onto the 2d screen. Every 3d object on the view is converted to 2d coords on the screen with a slight adjustment for height depending on the height of the object.

Don't fully understand the shader refactor, Liam should be asked if needing information on it.
## Impact

Will increase performance by not having to render a 3d texture to display the name of an object.

`createMeasurementLabel` was already not working due to the texture not being sampled in the shader.

Downstream implementations of `TextObject` will have to be refactored to work with the 2d overlay or keep it and implement texture sampling.
## Validation

Create a part with a name and enable `Show part names in view` from the options bar.


liamnwhite1 (changes requested):
Thank you for identifying this issue, @jkwookee. Your PR does fix the visibility problem, and nice work tracking down the specific commit that introduced the regression.

As you noted in the PR, this fix may add a small rendering performance cost. I think this is a good opportunity to replace the old approach with something better instead of restoring the previous behavior exactly.

Right now, part labels use `TextObject`, which creates a texture for the text. That works, but it is more complicated than we need and can add rendering overhead. A better long-term solution is to draw the label text with `QPainter` after the OpenGL scene is rendered. This lets us render readable labels without creating a separate text texture.

I pushed three commits to your branch to help get this started:

1. The first commit refactors the shaders to make the rendering code easier to read. This isn't relevant for the implementation but I thought it was as good a time to clean it up as any.
2. The second commit adds virtual overlay functions to `BaseView`, so views can draw 2D content after the 3D scene.
3. The third commit removes the part-name `TextObject` path and sets up the new overlay structure in `PartView`.

The next step is to implement the actual label drawing inside `PartView::paintOverlay()`. This will be more involved than the original fix, but it should give us a cleaner and more efficient solution and I believe you can do it!

Once that is working, please also try an interactive rebase to remove the first two commits from your original fix, since the final version should not rely on restoring the old texture-based label behavior. This will be a good opportunity to practice rebasing. Please let me know if you have any questions.

liamnwhite1 (approved):
This looks great! The only thing I noticed was dark-mode part labels can still be hard to read. In part_view.cpp (line 124) dark mode used Qt::darkGreen and drew text directly over the OpenGL scene with no backing fill/outline. The dark theme background is already dark in theme_tool.cpp (line 20), so this undermines the PR’s main “show part names” visibility fix for dark-mode users. I added a small fix to render part name overlays on a small theme-aware translucent background instead of relying on a colored text pen. This keeps labels readable in dark mode and when they overlap similarly colored parts. If this looks good to you, you can go ahead and merge!

AshGannon (approved):

jkwookee:
Looks good to me, merging. Was going with dark green to match the right side UI but this looks much better for both dark and light mode by having the label box.

PR #273, opened 2026-09-01
Add connected spiral insets option

aroschli:
## Summary
- Add a perimeter checkbox, `Connect to spiral insets`, shown when spiral perimeters and spiral insets are enabled.
- Allow spiral perimeter optimization to append computed spiral inset loops and skip the separate inset optimization pass when the option is enabled.
- Preserve the 45-degree forward connector behavior and apply inset segment settings after the perimeter-to-inset bridge.

## Validation
- `nix develop --command cmake --build build/generic-llvm-ninja --target ornlslicer_spiral_path_tests`
- `nix develop --command ctest --test-dir build/generic-llvm-ninja -C Debug -R spiral_path_tests --output-on-failure`
- Sliced a temporary copy of `ConnectedInsetsTest.s2p` with `inset`, `spiral_inset`, and `spiral_perimeter_connect_to_insets` enabled; verified the first perimeter-to-inset handoff had no travel/extruder-off between moves and measured 45.002 degrees relative to the incoming tangent.
- `git diff --check`

### DIFF
--- a/include/geometry/spiral_path.h
+++ b/include/geometry/spiral_path.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <algorithm>
+#include <array>
+#include <cmath>
 #include <limits>
 
 #include <qcontainerfwd.h>
@@ -13,9 +15,28 @@
 namespace ORNL {
 namespace SpiralPath {
 namespace detail {
+struct RayLoopIntersection {
+    Point point;
+    double distance;
+    int segment_index;
+};
+
 inline Point pointAlongSegment(const Point& start, const Point& end, double ratio) {
     return Point(start.x() + ((end.x() - start.x()) * ratio), start.y() + ((end.y() - start.y()) * ratio),
                  start.z() + ((end.z() - start.z()) * ratio));
+}
+
+inline double cross2D(double lhs_x, double lhs_y, double rhs_x, double rhs_y) {
+    return (lhs_x * rhs_y) - (lhs_y * rhs_x);
+}
+
+inline bool normalize2D(double& x, double& y) {
+    const double length = std::sqrt((x * x) + (y * y));
+    if (length <= std::numeric_limits<double>::epsilon()) { return false; }
+
+    x /= length;
+    y /= length;
+    return true;
 }
 
 inline Point stopPointOnClosingSegment(const Polyline& line, Distance distance_before_start) {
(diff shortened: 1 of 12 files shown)


aroschli:
Pushed an update in `1c39e8bb` that adds a new **Branch after tip wipe** checkbox for connected spiral insets.

Summary:
- Added `spiral_perimeter_connect_to_insets_after_tip_wipe` setting, defaulting off and shown when connected spiral insets plus perimeter tip wipe are enabled.
- When enabled, perimeter spiral paths run their modifiers/tip wipe first, then the connected inset spiral chain is appended from the wipe endpoint.
- Existing connect-to-insets behavior is preserved when the new checkbox is off.

Verification:
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_spiral_path_tests`
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer`
- `nix develop -c ctest --test-dir build/generic-llvm-ninja -C Debug -R '^spiral_path_tests$' --output-on-failure`

aroschli:
Pushed follow-up `1740eb4f` for the branch-after-tip-wipe feedback.

Updates:
- Moved perimeter **Branch after tip wipe** before **Connect to spiral insets** and made it a general spiral perimeter option.
- Relaxed **Connect to spiral insets** UI dependencies so it can be enabled when perimeter/inset/spiral perimeter are enabled, without requiring `spiral_inset` to already be checked in the UI.
- Added matching **Branch after tip wipe** for spiral insets.
- Perimeter and inset spiral branching now emit the branch after modifiers/tip wipe when the option is enabled; connected spiral insets use the same post-wipe branch point from the final perimeter path.

Verification:
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer_spiral_path_tests`
- `nix develop -c ctest --test-dir build/generic-llvm-ninja -C Debug -R '^spiral_path_tests$' --output-on-failure`
- `nix develop -c cmake --build build/generic-llvm-ninja --target ornlslicer`

aroschli:
Pushed follow-up `0a0a982b` to fix the inter-loop travel in branch-after-tip-wipe mode.

The branched perimeter/inset loops are now emitted as one continuous `Path`: each loop completes its tip wipe, then a normal extruding line connects that wipe endpoint to the next loop. Only the initial move into the first loop remains a travel. This also avoids writer-level path teardown/startup between branches.

Added `spiral_branch_tests` covering both perimeter and inset behavior. It verifies one continuous path, exactly one initial travel, geometric continuity, and a printing branch immediately after tip wipe.

Verification:
- full `ornlslicer` build
- `spiral_path_tests`
- `spiral_branch_tests`

aroschli:
Fixed the backward post-tip-wipe branch behavior in `f35c0105b`.

Branch planning now adjusts the preceding loop's seam against the next loop start and evaluates the connector from the actual forward tip-wipe endpoint. This preserves the loop's travel direction, including clockwise paths, for both spiral perimeters and spiral insets. It handles both partial-loop and **Complete path before connecting** modes.

Regression coverage now uses nonzero bead width/wipe distance, explicitly exercises clockwise paths, and asserts that the post-wipe branch remains continuous, contains no inter-loop travel, and has a positive forward dot product.

Verified with:
- full Debug `ornlslicer` build
- `spiral_path_tests`
- `spiral_branch_tests`

aroschli:
Follow-up fix pushed in `ab82c1086` for the long/erratic branch connections.

The previous seam selector prioritized a 45-degree branch angle, which could choose a distant point when densely segmented curves rejected the ideal interior candidate. Branch selection now:

- treats forward motion as a hard eligibility constraint
- ranks every eligible seam by connector distance first
- places interior seams just on the forward side of the nearest projection
- does not retain an already-forward seam when a closer forward seam exists

I updated the regression to use five densely segmented concentric circular loops matching the reported shape. Before this fix it reproduced branch lengths of about 130 and 91 units on loops spaced 5 units apart. The test now requires every branch to remain within 6 units while retaining a positive forward component, for clockwise perimeter/inset paths in both partial and completed-loop modes.

Verified with the full Debug application build plus `spiral_path_tests` and `spiral_branch_tests`.

aroschli:
Added local angled branch preference in `6fe39f967`.

The branch selector now evaluates both the nearest forward seam and an approximately 45-degree seam. It chooses the angled branch only when:

- its tangent is near 45 degrees
- it remains local (no more than 1.5× the nearest valid connector)

If an angled seam cannot satisfy the segment/minimum-length constraints, it falls back to the nearest near-orthogonal forward connection. This preserves the prior fix that eliminated long cross-model chords.

Coverage now verifies:
- angled connections relative to both the departing and joining bead tangents
- local connector length on the five-ring circular case
- explicit near-orthogonal fallback when no local angled seam fits
- perimeter and inset behavior, with partial and completed loops

Verified with the full Debug application build plus `spiral_path_tests` and `spiral_branch_tests`.

alexarbogast:
I tested the `Enable Spiral Inset` feature with `Branch after tip wipe` disabled. In my limited test cases, everything has the target functionality. 

Project: [connected_insets_no_tip_wipes.zip](https://github.com/user-attachments/files/32580650/connected_insets_no_tip_wipes.zip)

---

I did run into a strange behavior with `Branch after tip wipe` enabled. With the project included below, there is always one connected inset that is perpendicular to the subsequent path. It seems like a 45 deg connection should fit here, so I wanted to double check that this is the intended functionality.  I have added a green box around this connection in the image below.

Project: [connected_insets_tip_wipes.zip](https://github.com/user-attachments/files/32580768/connected_insets_tip_wipes.zip)

Screenshot from 2026-09-23 15-54-39

alexarbogast:
Also, could we add a separate bead type for the connections? This would help distinguish perimeters/insets vs the connections, since these are often a bit cluttered in the visualization.

aroschli:
Addressed both review comments in 650a76cc5:

- Angled connections: added a follow-up seam-selection pass for post-tip-wipe branches. I reproduced the attached 200 x 200 square case, and the available branch connections now use angled transitions instead of the perpendicular fallback.
- Separate connection type/color: branch moves now use a Spiral Connection path modifier. G-code comments identify them as PERIMETER SPIRAL CONNECTION or INSET SPIRAL CONNECTION, and they have a distinct gold default color that is configurable in Preferences. Keeping this as a modifier preserves the underlying perimeter/inset process settings.

Validation:
- ornlslicer_spiral_path_tests passes
- ornlslicer_spiral_branch_tests passes, including the reproduced square regression case
- The affected core, writers, loader, and Preferences code rebuild successfully
- The new visualization assertions pass. The visualization test still reports the same three pre-existing helical default-color mismatches present on develop.

aroschli:
Implemented the review follow-ups in commit 13d00b88b:

- only suppress the standalone inset region after the perimeter confirms connected inset segments were emitted, preserving insets in global spiralize and other early-return cases
- keep different-material perimeter/inset regions separate so region-level multi-material transition handling remains intact
- apply inset slowdown, tip-wipe, and spiral-lift settings to connected inset tails, with distances bounded to the inset portion
- limit post-tip-wipe branching to forward tip wipes in both runtime behavior and settings visibility; reverse/angled wipes retain the normal spiral path behavior
- require spiral insets before exposing the connect-to-insets option
- add island-level regression coverage for connected output, inset terminal modifiers, global spiralize fallback, different-material fallback, and reverse-wipe behavior

Verification:
- main Debug application target builds successfully
- spiral_path_tests passes
- spiral_branch_tests passes with the new integration cases
- git diff --check passes

The existing visualization_colors_tests still reports the same three helical default-color expectation mismatches present on develop; this change does not touch those defaults.

aroschli:
Implemented the review fixes in 845992860:

- Split branch-after-tip-wipe output at unrelated or non-local loops, inserting travel instead of a long extruding connector.
- Keep continuing branch connections at layer Z by suppressing wipe lift and spiral lift until the terminal path.
- Apply inset-specific completion and minimum-segment settings when insets are owned by the perimeter; clarified in the settings UI that the perimeter branch option controls a connected chain.
- Track connected-inset consumption per contour so partial emission cannot suppress the remaining inset region.
- Added regressions for mixed completion policies, separated geometry, lifted wipes, single/multi-perimeter connected handoffs, and complete connected-inset emission.
- Rebased the branch onto the current develop tip; the post-rebase tree is identical to the tested tree.

Validation:
- spiral_path_tests: passed
- spiral_branch_tests: passed
- Settings regenerated and both generated JSON files validated with jq
- git diff --check: clean

aroschli:
Resolved both review findings in a50a7920:

- Post-wipe perimeter-to-inset handoffs now require nested geometry, a forward branch, and a bounded connector. The receiving inset seam prefers a 45-degree intersection; unsafe handoffs are emitted as separately traveled inset paths instead of extruding across components.
- The perimeter-to-inset bridge now retains perimeter process settings. Inset width, speed, extrusion, and region settings begin on the first actual inset contour segment.
- Added regressions for a separated component with no corresponding inset and for the bridge region settings.

Validation:
- `ninja -C build/generic-llvm-ninja -f build-Debug.ninja ornlslicer`
- `ctest --test-dir build/generic-llvm-ninja -C Debug -R "^(spiral_path_tests|spiral_branch_tests)$" --output-on-failure` (2/2 passed)
- `git diff --check`

aroschli:
Applied the review follow-ups in commit f1c2ac67a:

- keep slowdown lift at zero when a perimeter or inset continues into a printing branch
- select before/after path hooks from the first/last printing segment and apply inset startup modifiers to standalone connected-inset fallback paths
- replace repeated QVector moves with single standard rotations, removing the cubic seam-rotation behavior
- add regressions for branch Z continuity, connected-inset startup, and writer hook regions

Validation: ornlslicer_spiral_path_tests and ornlslicer_spiral_branch_tests both pass (2/2); git diff --check is clean.

aroschli:
Implemented the review follow-ups in 703e6b14:

- Split continuous mixed perimeter/inset output into contiguous writer path and region scopes, so process schedules and per-path setup transition at the actual segment boundary without adding travel.
- Preflight the forward-wipe connection before suppressing terminal modifiers. Unsafe inset connections now keep the perimeter lift and fall back to travel.
- Added regression coverage for mixed-region writer hooks and unsafe traveled-inset lift preservation.

Validation:
- ornlslicer_spiral_branch_tests: PASS
- ornlslicer_spiral_path_tests: PASS
- ornlslicer_visualization_colors_tests: still reports the same 3 unrelated default-color assertion failures seen at the merge base.

aroschli:
Addressed the review findings and pushed commit 398b8a90.

- Rebased the branch onto current develop and resolved the visualization-color test conflict while preserving the SpiralConnection coverage.
- Connected inset bead width and adaptive settings now come from the localized segment SettingsBase, so SettingsPolygon overrides are retained in generated connected-inset paths.
- Added an end-to-end regression test that verifies a local inset bead-width override is consumed once and applied to every connected-inset printing segment.
- Updated the PR spiral-path tests to use the current ORNL::Testing helpers after the develop-side test utility refactor.

Validation:
- visualization_colors_tests: passed
- spiral_path_tests: passed
- spiral_branch_tests: passed
- clang-format strict check and git diff --check: passed

aroschli:
Implemented the requested review fixes in d14b28af:

- Preserve one writer path lifecycle across connected perimeter → inset region transitions, while still emitting the region transition hooks. This prevents `writeAfterPath()` from stopping deposition mid-path.
- Build spiral and branch geometry as open paths from the outset, including minimum-segment cleanup and localized-settings clipping, instead of creating a synthetic closing edge and deleting only the last generated segment.
- Strengthen `spiral_branch_tests` to cover one before/after path hook pair for mixed-region paths and completed branches with nonzero minimum-segment cleanup.

Validation:
- Full Debug build passed.
- `ctest --test-dir build/generic-llvm-ninja -C Debug --output-on-failure`: 22/22 tests passed.

aroschli:
Resolved the review findings in `15594f65`:

- Added an in-path region-transition hook so connected perimeter/inset paths keep continuous deposition while updating writer state. Cincinnati now switches to the inset acceleration without replaying path-start custom G-code, and Arc Specialties updates its region labels.
- Routed the single-perimeter + connected-inset + branch-after-tip-wipe case through the branch builder, so `Complete path before connecting = false` leaves the perimeter open as configured.
- Computed branch-loop emission eligibility before connection flags, preventing a filtered successor from suppressing terminal tip-wipe/lift modifiers on the preceding loop.
- Added regressions for mixed-path writer transitions, the one-perimeter open-loop handoff, and filtered branch successors.

Validation:

```text
ctest --test-dir build/generic-llvm-ninja -C Debug -R "^(spiral_path_tests|spiral_branch_tests|path_modifier_tests|arc_specialties_parser_tests)$" --output-on-failure
```

Result: 4/4 tests passed.

PR #282, opened 2026-09-10
Add Arc Specialties helical OptStopMode support

liamnwhite1:
## Summary

Adds Arc Specialties helical layer-start handling so each helical layer emits `V.E.OptStopMode = <number>` immediately after `;BEGINNING LAYER: ...`.

## Related issues

- No related issues.

## Details

This feature emits Arc Specialties `OptStopMode` from the actual post-ordering C-axis rotation direction:

- `V.E.OptStopMode = 1` for positive CP rotation.
- `V.E.OptStopMode = 2` for negative CP rotation.

The implementation stays scoped to `ArcSpecialtiesWriter`; no Arc Specialties controller behavior is added to `CylindricalLayer`. To support `Next Closest` ordering, the writer defers helical layer-local prefix output until the first ordered print move is known. It then computes the signed CP delta from that actual print move, emits `OptStopMode` immediately after the layer marker, and releases the buffered travel/setup output and deferred `G80` schedule in the correct order.

This also preserves Arc Specialties block-number behavior by keeping `OptStopMode` unnumbered while continuing to number the deferred `G80` schedule and executable setup/print lines.

## Impact

This changes Arc Specialties helical G-code output by adding one layer-scoped controller assignment near the top of each helical layer. The emitted value now reflects the direction of the first post-optimized helical print move, including paths reversed by cylindrical `Next Closest` ordering.

Risk: Low. The change is isolated to the Arc Specialties writer and focused parser/writer tests. Non-Arc Specialties writers and generic cylindrical layer ordering are not changed.

## Validation

Validated with: 
[Donut3.1.rev-10.zip](https://github.com/user-attachments/files/32063918/Donut3.1.rev-10.zip)

PR #283, opened 2026-09-11
Add Helical Tool Start Angle Offset for Arc Specialties Output

liamnwhite1:
## Summary

Adds `Helical Tool Start Angle Offset`, a signed Arc Specialties tool/CP baseline offset for helical cylindrical slicing.

This replaces the older absolute helical start-angle semantics with a clearer tool-offset setting: emitted helical X/Y geometry still starts at the generated top-dead-center start point, while Arc Specialties `CP` and helical `XR`/`YR` output are shifted by the effective tool offset. The change also migrates legacy saved settings, updates bundled templates and documentation, and expands focused regression coverage for writer/parser output.

## Related issues

- Depends on #282 

## Details

This PR changes the helical start-angle setting from an absolute geometry start angle into a tool-frame/positioner offset:

- Renames the setting to `helical_tool_start_angle_offset` / `Helical Tool Start Angle Offset`.
- Keeps the generated helical X/Y path start anchored at top dead center (`+Y`) when the offset is `0 deg`.
- Applies the effective offset to Arc Specialties helical `CP`.
- Applies half of the effective offset to helical `XR` and `YR`.
- Preserves signed, unwrapped helical `CP` behavior so left-handed and right-handed helical output can reflect the actual angular sweep.
- Mirrors the effective offset sign for alternating helical passes when path ordering reverses the radius pass direction.

Implementation updates include:

- A dedicated helical tool start angle helper in `include/slicing/helical_tool_start_angle.h` and `src/slicing/helical_tool_start_angle.cpp`.
- Cylindrical slicer logic that stores the effective per-layer helical tool start angle offset on layer settings.
- Arc Specialties writer logic that uses the effective offset consistently for:
  - initial world approach
  - helical travel
  - travel-lower moves
  - helical print moves
  - header comments
- Settings constants, canonical YAML metadata, generated config files, and bundled `.s2c` templates updated to the new setting key and display name.
- Settings schema bumped to `12.0`.
- Migration support for both:
  - older `helical_path_start_angle` absolute values, converted to the new tool offset semantics
  - intermediate `helical_start_angle_offset` values, migrated directly to `helical_tool_start_angle_offset`
- Legacy G-code footer import migration coverage for the renamed setting.
- Arc Specialties and cylindrical slicing docs updated to clarify that this offset changes the tool/CP baseline, not the emitted X/Y helix phase.

The requested five-commit range also includes a small `.codex/AGENTS.md` cleanup that removes redundant local guidance about keeping diffs narrow.

## Impact

Functionality impact: Medium.

Users now configure helical Arc Specialties tool alignment with a signed tool start angle offset instead of an absolute geometry start angle. This better matches the actual behavior needed by the controller: the path remains geometrically anchored at the generated helical start, while `CP`, `XR`, and `YR` carry the machine/tool offset.

Workflow impact:

- Existing saved projects and imported G-code footers using legacy helical start-angle keys are migrated forward.
- New projects and templates use `Helical Tool Start Angle Offset`.
- Documentation now distinguishes helical path geometry from Arc Specialties tool/positioner orientation.

Compatibility impact:

- Legacy `helical_path_start_angle` values are migrated as absolute-angle values into the new offset semantics.
- Intermediate `helical_start_angle_offset` values are migrated directly to the final setting key.
- Generated configs and bundled templates are updated with the current setting name.

Risk level: Medium.

The change affects machine-output semantics for helical Arc Specialties paths, including `CP`, `XR`, and `YR` fields. The risk is reduced by keeping X/Y helix geometry anchored, adding explicit migration paths, and expanding focused regression coverage around the affected G-code output.

## Validation

Validation coverage added or updated in this PR includes:

- Arc Specialties parser/writer expectations for helical `CP` baseline behavior.
- Regression coverage that verifies the helical tool offset is preserved through writer output.
- Coverage for applying half of the effective offset to helical `XR` and `YR`.
- Output checks for helical world approach, travel, travel-lower, print moves, and header comments.
- Settings migration tests for legacy `helical_path_start_angle` and intermediate `helical_start_angle_offset` values.
- G-code footer import migration coverage for legacy helical start-angle settings.
- Helical clipping/rounding expectations updated for the new effective offset helper.

[Donut3.1.rev-999.zip](https://github.com/user-attachments/files/32125047/Donut3.1.rev-999.zip)


liamnwhite1:
**Findings**
- [P1] Offset sign is chosen before the optimizer may reverse the printed helix. [cylindrical_slicer.cpp](/home/fnx/develop/ornl-slicer-helical-tool-offset/src/threading/slicers/cylindrical_slicer.cpp:804) stores an effective `helical_tool_start_angle_offset` using only generated radius-pass parity, but the actual print direction is decided later in `PathOrderOptimizer::linkNextHelicalPath()`, which can reverse a path when entering from its end. See [path_order_optimizer.cpp](/home/fnx/develop/ornl-slicer-helical-tool-offset/src/optimizers/path_order_optimizer.cpp:350). This breaks cases where `Complete Revolution`/`Next Closest` skips an intermediate radius due to clipping: the next retained radius can still be reversed by endpoint proximity, but its parity-derived offset is not flipped. That makes CP/XR/YR use the wrong sign for that printed pass. The sign should follow the post-ordering direction, or the layer/segment settings should be adjusted when the optimizer reverses the helical path.

PR #286, opened 2026-09-16
feat: add rename part option

### LINKED ISSUE #12: Corner Ramping Based on Path Type

### LINKED ISSUE #281: Add ability to rename part
## Summary

Users should have the ability to rename a part from within Slicer.

## Problem or opportunity

Currently, users do not have the ability to rename parts from within Slicer and so as project names change, the underlying part does not. This results in confusion when trying to version things.

## Proposed capability

Users should have the ability to rename parts from within the Slicer gui.

## Acceptance criteria

- The user should be able to select a part in the scene and be presented with an option to "Rename Part" in the part's context menu.
- The user should be able to select a part in the part list and be presented with an option to "Rename Part" in the part's context menu.
- When the user selects the "Raname Part" option from the part's context menu, a window should appear with an appropriate prompt, a text input for the new part name, and an "OK" button.
- When the user enters a new name for the part and clicks the "OK" button, the "Rename Part" window should disappear, and the new part name should be shown in the part list.
- When a part name is changed and a project is saved, the project should save the part with the new name.

jkwookee:
## Summary

Adds a new utility file for dialogue related methods, moving `promptForName` to there.

Adds a option in the context menu to rename a part.

Refactors `right_click_menu` to have the code ordered by the order they appear in the context menu.
## Related issues

- Closes #281 
## Details

Creates a new event to update the tree in `part_control` when a part rename is successfully done in `right_click_menu`

Refactors `right_click_menu` to have each possible actions ordered by the way they appear in the context menu so that there is a consistent pattern to where you can find code for a specific context menu option.
## Impact

From issue:

> Currently, users do not have the ability to rename parts from within Slicer and so as project names change, the underlying part does not. This results in confusion when trying to version things.

## Validation

Build and use the right click context menu on a part to access the rename option. Rename a part and observe the part name changing in the tree. Check the name in the info option of the right click context menu. Save project then reload the project file to see the name being kept.


liamnwhite1 (changes requested):
Thanks for putting this together. The featurelooks great and is in the right place UX-wise, but I think we need to fix a few state-handling issues before merging.

1. Renaming a part only updates `Part::m_name`, but `SessionManager::m_parts` is still keyed by the old name. That leaves the app in an inconsistent state: after a rename, deleting the part removes it from the tree, but `SessionManager::removePart(item->part())` can return false because it looks up the new name in a map still keyed by the old name. This can leave the renamed part in the session and affect slicing/saving/duplicate-name checks. Please route renames through a session/model method that updates the session map key and the `Part` name together.

2. Canceling the rename dialog currently clears the part name. `DialogUtils::promptForName(...)` returns an empty string when the user cancels, and the rename action calls `item->renamePart(new_name)` unconditionally. Please skip the rename when the dialog is canceled or returns an empty name.

3. A rename does not mark the project modified. `PartMetaModel::nameUpdate` is emitted, but `MainWindow` only marks the project dirty for add/reload/remove/parenting/transform updates. Please connect `nameUpdate` to `markProjectModified()` too, so users get the expected save prompt after renaming.

When finished, please manually verify: rename a part, cancel a rename, rename then delete, and rename then close/save/reopen.

jkwookee:
> Please route renames through a session/model method that updates the session map key and the `Part` name together.

~~Do you want the part rename to be done through `session_manager` as well rather than `part_meta_item`? Currently addressing reviews to have only the key renaming done through `session_manager` with a generic key rename method and keeping renaming the part and emitting the tree update through `part_meta_item`~~

Having rename and key update in one method instead

liamnwhite1:
@jkwookee Thanks for addressing the earlier rename issues. One small edge case is still present:

When renaming a part, the dialog pre-fills the current name, but `DialogUtils::promptForName()` still treats that same name as a duplicate because `CSM->getPart(name)` returns the current part. So if a user opens Rename Part and presses OK without changing the text, they get “Name already in use” even though keeping the existing name should be valid.

Could you update the duplicate check to allow the unchanged current name please? One option is to pass the current `Part` into the helper and only reject matches owned by a different part.

jkwookee:
Edge case should be addressed, tested and it allows for the same name to be inputted and selected with Ok but won't trigger an actual rename

liamnwhite1 (approved):
These fixes look good! Thank you for the hard work!

liamnwhite1:
@jkwookee Please hold on merging this for a moment. I need another PR to get in first. I'll merge this when it's time.

PR #287, opened 2026-09-18
feat: make "Use Variable for Z Position" be a dropdown

### LINKED ISSUE #12: Corner Ramping Based on Path Type

### LINKED ISSUE #285: Create Hi-Temp Variable Z Checkbox / Dropdown
Currently, there is a checkbox for 'use variable for Z position' which allows for a machine variable (#200) to be used as the Z offset. This works perfectly, but for high temp materials, that value needs to be modified due to the growth of the extruder at higher temperatures, and then reverted after the print. This is a request to change that checkbox to a dropdown (or maybe just add a second checkbox) to distinguish between using a low-temp Z offset (#200) and a high temp Z offset (#201)

jkwookee:
## Summary

Changes Use Variable for Z Position toggle in Dimensions to be a drop down between three options of nothing, variable 200, and variable 201.
## Related issues

- Closes #285 
## Details

From issue:
> This works perfectly, but for high temp materials, that value needs to be modified due to the growth of the extruder at higher temperatures, and then reverted after the print. 

## Impact

Creates another migration for the setting. Migrations has a bit of boilerplate that could be reduced / refactored cause as more migrations are added the `settings_version_control.cpp` could balloon a lot.

Could make `Use Variable for Z Position` setting into a input for a specific desired variable number but I'm not sure if that is desired / keeping it specific is better?

## Validation

Build then find the dropdown option in Dimensions while Machine Setup has the Syntax setting set to Cincinnati. Slice a test object with the setting set to either 200 or 201 and find `#200` or `#201` in the resulting g-code.


aroschli:
Suggested changes from review:

1. Update the importer/version-control test expectations for the new master version.

   `SettingsVersionControl::rollSettingsForward()` now includes the `pre_13_0To13_0` migration, so `tests/gcode_settings_importer_tests.cpp` should no longer expect rolled settings to stop at `12.0` around lines 80-85. Those assertions should move to `13.0`, and this would be a good place to also assert the new `variable_for_z` bool-to-enum migration.

2. Preserve compatibility when importing old G-code settings footers.

   Older generated G-code footers can contain `;variable_for_z true` or `;variable_for_z false` because `WriterBase::writeSettingsFooter()` writes raw JSON values. The `.s2c` version migration handles this, but `GcodeSettingsImporter::importFile()` validates footer values directly against current `master.conf`. Since `variable_for_z` is now an `enumeration`, `validateValue()` rejects JSON booleans with `variable_for_z must be an enumeration index.` Consider converting this specific legacy boolean in the footer importer path, or teaching enum validation to accept booleans for this migrated setting only.

Verification: `git diff --check develop...HEAD` passed. I started a targeted build of `ornlslicer_gcode_settings_importer_tests`, but the local machine crashed before the binary/test run completed, so I do not have a completed test result.

jkwookee:
Should be ready for re-review, updated the PR description with the test refactor/fix changes made. Added the 12.0 to 13.0 test for `variable_for_z` in a new separated file specifically for `s2c` migration. For preserving gcode compatibility I went with changing `migrateBooleanToInt` to be compatible with updating gcode settings and then added `migrateBooleanToInt` to `migrateLegacySettingKeys`.

aroschli:
**Code Review**

No blocking findings. The bool-to-enum migration, raw G-code footer import path, and Cincinnati writer mapping all look coherent.

**Residual Risk**

There is not direct automated coverage for selecting `#201` and asserting Cincinnati output uses `#201`; the mapping lives in `src/gcode/writers/cincinnati_writer.cpp`, while the new tests cover legacy `true` migrating to `#200` in `tests/gcode_settings_importer_tests.cpp`. Not a blocker, but a small writer-level test would pin the new option down nicely.

**Verification**

Built the full Debug test target set, then ran:

`ctest --test-dir build/generic-llvm-ninja -C Debug --output-on-failure`

Result: 21/21 tests passed. Tracked worktree is clean.

aroschli (approved):

jkwookee:
Merging in a bit if there are no concerns, other PRs that are implementing tests should be refactored to use `test_utils` for similar methods

PR #293, opened 2026-09-21
fix(slicing, gcode): use surface-referenced Z for Wire Arc paths

liamnwhite1:
## Summary

Correct Wire Arc deposition heights so planar paths reference the surface receiving the deposited bead rather than the completed top of that bead.

For a constant 2.8 mm layer height, Wire Arc output now follows `Z0.0000`, `Z2.8000`, `Z5.6000`, and so on. The change also corrects Wolf’s initial approach move so travel lift is measured from the actual first target and the printer Z offset is applied exactly once.

## Related issues

- No related issues.

### Bug reproduction and root cause

The supplied project uses:

- Wire Arc machine type
- Wolf G-code syntax
- 2.8 mm layer height
- 0 mm printer Z offset
- 12.7 mm travel lift

Previously, its first deposition path was emitted at Z2.8 rather than Z0.0.

Planar cross-sections are intentionally calculated at the midpoint of each layer. During coordinate restoration, the generic layer logic then added another half-layer minimum-Z shift. For a horizontal 2.8 mm layer, those two operations produced:

- 1.4 mm cross-section midpoint
- 1.4 mm minimum-Z restoration
- 2.8 mm first deposition coordinate

That convention is appropriate for toolpaths representing the completed top of a layer, but Wire Arc paths need to represent the existing surface onto which the next bead is deposited.

Wolf’s initial travel also constructed an absolute height from the Z offset, travel lift, and layer height before `writeCoordinates()` applied the Z offset again. This coupled machine calibration to the layer datum and could apply a nonzero offset twice.

### Implementation

- Added an explicit Wire Arc deposition-surface reference in `Layer`.
- Retained midpoint cross-sectioning while shifting restored Wire Arc paths down half of the effective layer height along the normalized slicing-plane normal.
- Bypassed the generic top-of-layer minimum-Z shift for Wire Arc paths.
- Used each layer’s effective height, preserving the correct accumulated surface for variable-height slicing.
- Preserved the existing top-of-layer behavior for ordinary non-spiral paths on other machine types.
- Changed Wolf’s first travel destination to the first path target plus the configured travel-lift vector.
- Kept `z_offset` as machine calibration applied by coordinate formatting exactly once.
- Normalized values below half of Wolf’s four-decimal coordinate resolution so numerical zero is emitted as `Z0.0000`, not `Z-0.0000`.
- Added a focused CTest target covering:
  - the fixed-height `0.0, 2.8, 5.6` Wire Arc sequence;
  - unchanged non-Wire Arc behavior;
  - variable layer heights;
  - angled slicing-plane normals;
  - Wolf first-approach clearance;
  - nonzero Z-offset application;
  - exact zero formatting.
- Updated the slicing and G-code architecture documentation with the new coordinate contracts.

### Existing branch content

The current `origin/develop...HEAD` comparison also includes two pre-existing documentation commits that:

- add parallel Debug build instructions to `docs/wiki/Getting-Started.md`;
- remove an optional GitHub Copilot mention from that guide.

These changes predate the WAAM fix and are not required by it.

## Impact

Wire Arc planar G-code now uses the deposition-surface convention:

- layer 1: Z0.0000
- layer 2: Z2.8000
- layer 3: Z5.6000

This intentionally changes existing Wire Arc output downward by one layer height relative to the previous convention. Non-Wire Arc output remains unchanged.

Projects that used a negative Z offset equal to the layer height as a workaround must remove that workaround and restore the actual machine-calibration offset. For the supplied project, `z_offset` should remain `0`.

Variable-height and angled planar slicing use the same surface-reference contract because the adjustment uses the current layer height and slicing-plane normal.

Risk level: **Medium**. The implementation is narrowly gated to the Wire Arc machine type, but it intentionally changes machine coordinates for existing Wire Arc builds. Automated and project-level slicing validation passed; physical controller and hardware execution were not performed.

## Validation

- Configured successfully with:

  `nix develop .#ornlslicerDev -L --command cmake --preset generic-llvm-ninja`

- Built the application, new WAAM regression target, and existing variable-layer-height target successfully.
- Ran:

  `nix develop .#ornlslicerDev -L --command ctest --test-dir build/generic-llvm-ninja -C Debug -R '^(waam_layer_z_tests|variable_layer_height_tests)$' --output-on-failure`

  Result: **2/2 tests passed**.

- Completed a full 410-layer slice of the supplied project:
  - first approach: Z12.7000
  - first deposition surface: numerical Z0
  - second deposition surface: Z2.8000
- The final regression explicitly verifies that numerical zero is serialized as `Z0.0000`.
- `clang-format --dry-run --Werror` passed for all modified C++ files.
- Both commits passed the repository pre-commit formatter.


liamnwhite1:
fe39130 addresses:
[P1] First travel can remain at clearance height — [wolf_writer.cpp (line 164)](/home/fnx/develop/ornl-slicer-waam-layer-z/src/gcode/writers/wolf_writer.cpp:164). The first travel unconditionally adds travel_lift, but the lowering move requires travel_lift_required. If the travel is shorter than minimum_travel_for_lift, that flag is cleared at lines 143–146, so the lowering move is omitted. The next deposition segment can therefore descend from clearance height while material is active. kNoLift has the same mismatch. The regression test fixes the threshold at zero ([waam_layer_z_tests.cpp (line 156)](/home/fnx/develop/ornl-slicer-waam-layer-z/tests/waam_layer_z_tests.cpp:156)), so it cannot catch this. The first approach and lowering decision should be paired, with coverage for a filtered short first travel.

PR #298, opened 2026-09-22
Fix global layer island order settings

aroschli:
## Summary
- use the active printing layer's settings for global-layer island ordering
- use the layer's custom island-order anchor and optimization frame instead of always reading global settings
- keep the global settings fallback when no printing layer is available

## Verification
- nix develop -c cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer -j2
- nix develop -c ctest --test-dir build/generic-llvm-ninja -C Debug -R optimization_anchor --output-on-failure
- manually verified the provided .s2p: UI layer 120 / slice layer 119 now starts with the outer perimeter bead when the layer-local custom optimization location is outside the object


aroschli:
Applied the follow-up for the global-layer island order review item.

What changed:
- `GlobalLayer` now uses per-layer island-order settings only when all orderable printing layers in the global layer agree.
- If island-order settings disagree, it falls back to `global_sb` and logs a warning instead of letting the lowest-sorted part UUID choose behavior.
- Custom island-order anchors now require a common layer frame; otherwise the code avoids projecting through an arbitrary layer frame.
- Added `global_layer_tests` covering both agreed layer settings and mixed-settings fallback, and registered it with CTest.

Verification:
- `nix develop --command ctest --test-dir build/generic-llvm-ninja -C Debug -R global_layer_tests --output-on-failure`
- `git diff --check`

aroschli:
Fixed the review finding about conflicting layer frames clearing the fallback anchor. The follow-up commit keeps the first orderable printing layer as the fallback anchor, so global custom island-order settings can still be projected when layer-local settings are rejected. Verification: rebuilt ornlslicer_global_layer_tests with nix develop, ran ornlslicer_global_layer_tests, ran ornlslicer_optimization_anchor_tests, and ran git diff --check.

liamnwhite1 (changes requested):
[P1] Do not treat per-part shifts as incompatible frames

Horizontal parts naturally receive different mesh-derived shifts, but for a +Z slicing plane the optimization-frame rotation is the identity, so those shifts produce the same custom anchor transform. This equality check therefore marks ordinary separated parts as conflicting and lines 113-117 discard their identical layer-local Custom Location settings in favor of global settings. A reproducer with both layers using custom X=2000 selected the near island via the global Next Closest fallback instead of the far island. Compare the effective projected anchor or transform rather than raw shift equality, and cover identical custom settings on horizontally separated parts.

The false conflict occurs because each part’s shift is derived from its own mesh midpoint in [cross_section.cpp (line 187)](/home/fnx/.codex/worktrees/ef83/ORNLSlicer/src/cross_section/cross_section.cpp:187), while the horizontal anchor transformation in [optimization_anchor.cpp (line 84)](/home/fnx/.codex/worktrees/ef83/ORNLSlicer/src/optimizers/optimization_anchor.cpp:84) is unaffected by differing X/Y shifts.

aroschli:
Addressed the requested frame-compatibility change in commit `99c7397d`.

What changed:
- Compare the effective projected custom island-order anchors instead of raw per-part shifts.
- Added a regression test for identical Custom X=2000 settings on horizontally separated +Z parts, confirming the layer-local Custom Location remains active.

Verification:
- `nix develop --command ctest --test-dir build/generic-llvm-ninja -C Debug -R global_layer_tests\|optimization_anchor_tests --output-on-failure`
- `nix develop --command clang-format --dry-run --Werror src/step/global_layer.cpp tests/global_layer_tests.cpp`
- `git diff --check`

aroschli:
Addressed the effective-anchor comparison issue in bd9067376:\n\n- compare each Custom Point layer's projected anchor using that layer's own settings and optimization frame\n- fall back to global settings only when island-order methods or effective anchors actually conflict\n- add regression coverage for differing raw custom Z values that project to the same horizontal-plane anchor\n\nVerification:\n- built ornlslicer_global_layer_tests and ornlslicer_optimization_anchor_tests (Debug)\n- both focused CTest tests pass\n- clang-format and git diff --check pass

PR #300, opened 2026-09-23
Fix seam attractor guide drag updates

aroschli:
## Summary
- carry the picked optimization point's attractor guide through drag state
- move the guide line with the dragged seam marker in both G-code and part views
- preserve the existing setting commit behavior on drag finish

## Verification
- cmake --build build/generic-llvm-ninja --target ornlslicer_obj --config Debug
- git diff --check


aroschli:
Reviewed PR #300. I did not find any blocking issues.

The change looks coherent: the picked optimization guide is carried into drag state and translated with the seam marker in both PartView and GCodeView. GraphicsObject::translateAbsolute() preserves the guide's rotation and scale, so the attractor direction/length should survive the move.

Verification run:
- git diff --check origin/develop...HEAD
- cmake --build . --config Debug --target ornlslicer_optimization_anchor_tests
- ctest -C Debug -R optimization_anchor_tests --output-on-failure
- cmake --build . --config Debug --target ornlslicer

Residual risk: I did not manually run the GUI drag flow in the live Part/GCode views.

PR #301, opened 2026-09-23
Fix weight estimate bead height inference

aroschli:
## Summary
- prevent tip wipe/startup/slowdown/coasting/ramping modifier moves from updating inferred bead height state
- keep modifier moves contributing volume using the current bead geometry
- add a Cincinnati parser regression for tip-wipe Z lift contaminating following perimeter volume

## Verification
- `build/generic-llvm-ninja/Debug/ornlslicer_common_parser_tests`
- re-sliced a representative large-format project locally and confirmed the generated header estimate now aligns with the path-volume calculation


aroschli:
Follow-up pushed in `7385d4cd`: modifier moves still do not update the bead-height inference anchor, but deposited modifier moves are now tracked as a fallback layer anchor. On layer reset, if the layer had deposition but no normal inference-eligible move, the last deposited position is promoted so the next layer does not infer across multiple layer heights.

I also treated `PRESTART` as a bead-height-inference modifier and added a Cincinnati parser regression for a modifier-only middle layer.

Verification:
- `cmake --build build/generic-llvm-ninja --config Debug --target ornlslicer_common_parser_tests`
- `build/generic-llvm-ninja/Debug/ornlslicer_common_parser_tests`
- `git diff --check`

aroschli:
Implemented the follow-up fix in 98dae477: modifier-only layers now keep the first non-inference deposition as the fallback bead-height anchor, so a later tip-wipe/Z-lift move does not anchor the next layer. Added a common parser regression for startup-at-layer-plane followed by tip-wipe lift before the next normal layer. Verification: ninja -C build/generic-llvm-ninja ornlslicer_common_parser_tests; build/generic-llvm-ninja/Debug/ornlslicer_common_parser_tests; git diff --check.

aroschli:
**Review result**

No blocking issues found in PR #301. The new bead-height inference flag is threaded through the linear and arc paths consistently, and the layer fallback-anchor lifecycle looks sound.

**Verified**
- `git diff --check origin/develop...HEAD`
- `ninja -C build/generic-llvm-ninja ornlslicer_common_parser_tests` reported the target up to date
- `build/generic-llvm-ninja/Debug/ornlslicer_common_parser_tests` passed

CI note: pre-commit and macOS/Ubuntu check jobs were green when checked; Linux/Windows jobs were still running. The failing “Rebase open develop PRs” job appeared unrelated to this PR: it reported a conflict rebasing `feat/connected-insets`.

